Issue #71, opened 2015-03-20
GitHub API rate limit exceed for  change log generation in large repos (> 3000 issues) 


skywinder:
Since I need to fetch details about each issue separately - and GitHub api has limit to 5000 requests per hour - Need to figure out some thing. 
Generate second token as temporary workaround?

phene:
Just ran into this myself on a large repo. Are there intermediary files created during the run so it can be resumed?

skywinder:
Unfortunately not yet.

sneal:
What about limiting the amount issues pulled back via a command line arg? I don't really need everything from 2 years ago, just the last few months would be great.

`github_changelog_generator --max-issues 500`

skywinder:
@sneal  Yep, it's a good idea! I will consider this in next release.

skywinder:
Ok, I added fallback for this error, so version `1.3.11` not crashes anymore - just print warning  message, that rate limit exceed!

estahn:
There are currently two options in the pipeline to fix this:
* Amending missing versions instead of regenerating all of them (#209)
* Use [octokit/octokit.rb](https://github.com/octokit/octokit.rb) with [Faraday HTTP Cache](https://github.com/octokit/octokit.rb#caching)

Example:
```ruby
project = 'skywinder/github-changelog-generator'

# Enable cache
stack = Faraday::RackBuilder.new do |builder|
  builder.response :logger if options[:debug]
  builder.use :http_cache, store: ActiveSupport::Cache::FileStore.new(File.join(Dir.tmpdir, 'ghclgen')) if options[:cache]
  builder.use Octokit::Response::RaiseError
  builder.adapter Faraday.default_adapter
end
Octokit.middleware = stack
# Octokit.auto_paginate = true

@client = Octokit::Client.new
releases = Octokit.releases project, :per_page => 2
releases.each do |release|
  puts release.inspect
end
tags = Octokit.tags project, :per_page => 2
tags.each do |tag|
  puts tag.inspect
end
issues = Octokit.issues project, :per_page => 2, :filter => :all, :state => :closed
issues.each do |issue|
  puts issue.inspect
end
```

daviddannunzio:
Is there an eta for when the `--max-issues` argument fix will be released?

skywinder:
@daviddannunzio next week for sure!

daviddannunzio:
Thanks @skywinder !

daviddannunzio:
[...]

PR #76, opened 2015-03-21
Add --max-issues argument to limit requests

sneal:
This addresses issue #71 by providing a configurable limit on the number of issues to retrieve. By default this limits the number of issues to 500 which should help users stay under the GitHub API limit and keep change logs from growing ridiculous in size for older repos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ Type `github_changelog_generator --help` for detailed usage.
         --[no-]compare-link          Include compare link between older version and newer version. Default is true
         --include-labels  x,y,z      Issues only with that labels will be included to changelog. Default is 'bug,enhancement'
         --exclude-labels  x,y,z      Issues with that labels will be always excluded from changelog. Default is 'duplicate,question,invalid,wontfix'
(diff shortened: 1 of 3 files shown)


skywinder:
@sneal hi, thanks for participating! It's a good addition, that make more flexibility in change log generation. I have a couple proposal before merge:

1. I think it's better to not limit users with 500 issues by default. (it depends of repo, but usually it can be work with more than 1000 issues). It's better to set ` --max-issues` to `nil` by default, and limit it only in case, if user explicit set this number.

2. description about this option in warning message, that I add in #75 )

3.  It will be nice to add mention about this option in **FAQ**  section, for example with topic:
"I got warning `Warning: GitHub API rate limit exceed (5000 per hour), change log may not contain some issues.` What it means and what I should do?"
And as answer:  add suggestion to run script with your key: ` --max-issues` 1 hour later (to reset rate limit), to get more predictive result.

skywinder:
@sneal Thanks, awesome work! :+1:

PR #207, opened 2015-03-23
Relatively require github_changelog_generator library

sneal:
During development its possible for the gem installed version of github_changelog_generator lib to get required instead of the current development version of the library. This ensures that Ruby pulls in the correct version of the lib, specifically for development without breaking production builds.


skywinder:
Thanks again, @sneal!

PR #216, opened 2015-03-25
Implement Reader class to parse ChangeLog.md

### LINKED ISSUE #212: Parsing of existing Change Log file
This is sub-task of #209.

As @estahn mention:

The `CHANGELOG.md` needs to be parsed and divided into sections which represent the versions, e.g. 
```ruby
    [{ :version => 'v0.0.1', :content => '...' }, { :version => 'v1.0.0', :content => '...' }, ...]
```

It's needed to implement:
- `--amend` #211
- `Ability to keep manual added information for older version tags` #213

estahn:
This will fix "Parsing of existing Change Log file". 
Fix #212.


skywinder:
@estahn you implement really handy module! :clap: 
Thanks! I will review it soon.

lib/github_changelog_generator/reader.rb
```diff
@@ -0,0 +1,74 @@
...
+    # - ## [v1.0.2](https://github.com/zanui/chef-thumbor/tree/v1.0.1)
+    # - ## v1.0.2 (2015-03-24)
+    # - ## v1.0.2
+    #
+    # @param [String] heading Heading from the ChangeLog File
+    # @return [Hash] Returns a structured Hash with version, url and date
+    def parse_heading(heading)
+      structures = [
+          /^## \[(?<version>v.+?)\]\((?<url>.+?)\)( \((?<date>.+?)\))?$/,
+          /^## (?<version>v.+?)( \((?<date>.+?)\))?$/,
```
skywinder: @estahn  Most of all project uses prefix "v" for version tags, but not all of them.
As example of big projects, that don't use:
- https://github.com/jquery/jquery
- https://github.com/bartaz/impress.js
- https://github.com/jashkenas/backbone

and this project also don't use any prefix for the tag versions.

So, I think it's needed to make char "v" optional in this regex. What do you think?

estahn:
@skywinder I changed the parser to allow any kind of version number. Also added a bunch of rspec tests and rubocop to check for validity.

Travis is currently failing because most of the files are not compliant with rubocop. You could just run `rubocop --auto-correct <filename>` to fix the issues automatically.

Thoughts?

skywinder:
@estahn thanks for rspec and fixes! Much appreciated!

Tell the truth I'm new in ruby. Ans it's my first serious module-based project. And you are motivate me to do it well with your commits!

- To not mix different tasks in one issue and make it clean - I will create separate PR with rspec integration. Please, merge it to your branch.
- Since you already add rspec to project, I think this parser - is good point to start implement tests! (I'm going to learn, how to write tests and implement first test case soon)
- As I can see, most of libs use module, named `options.rb` to implement options parse logic. I wrongly name in by `parser.rb`. But think that `parser.rb` is more suitable for your module. So, I think it will be better to rename  `parser.rb -> options.rb` and `reader.rb -> parser.rb`. What do you think?

skywinder:
Merged. I also mute all the rubocop's warnings, that can't be fixes automatically. I will eliminate them later one-by-one.

PR #220, opened 2015-03-26
Cleanup

tuexss:
I tried to add a bit of consistency and remove unused or unnecessary lines of code. I hope this helps a bit.

### DIFF
--- a/lib/github_changelog_generator.rb
+++ b/lib/github_changelog_generator.rb
@@ -292,12 +292,11 @@ module GitHubChangelogGenerator
           tags.concat(page)
         end
         print "                               \r"
-        if @options[:verbose]
-          puts "Found #{tags.count} tags"
-        end
 
         if tags.count == 0
           puts "Warning: Can't find any tags in repo. Make sure, that you push tags to remote repo via 'git push --tags'".yellow
+        elsif @options[:verbose]
+          puts "Found #{tags.count} tags"
         end
 
       rescue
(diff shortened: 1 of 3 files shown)


lib/github_changelog_generator.rb
```diff
@@ -139,7 +139,7 @@ def fetch_merged_at_pull_requests
       }
 
       if @options[:verbose]
-        puts 'Fetching merged dates... Done!'
+        puts "Fetching merged dates: Done!"
```
skywinder: Thanks for cleanup! I think there is should be single quotes, since you don't use any parameters.

tuexss: I would've gone with http://viget.com/extend/just-use-double-quoted-ruby-strings and always use double quotes. But if you prefer mixed style, I can revert this part :)

skywinder:
Wow! @tuexss it seems you are deeply inspect this code and fix bunch of different typos and logic issues! Now it looks much healthy! :gem: 
Thanks a lot! :+1:

lib/github_changelog_generator.rb
skywinder: @tuexss it really makes sense. I just new in ruby, and consider, if RubyMine and rubocop pronouncedly highlight this - it should be like that "a priori". So, I thinking to move to double quote and revert this "fancy" PR :)

.. and add to rubocom.yml.

    Style/StringLiterals:
      EnforcedStyle: double_quotes

Thanks, for point it out!

tuexss:
Yes, i'm also quite new in ruby. Thanks for the merge!

PR #221, opened 2015-03-26
Use thor instead of OptionParser

estahn:
[thor](http://whatisthor.com) is an (arguably) nicer  alternative to OptionParser. Have a look and let me know if you would like to pursue thor rather then OptionParser.

### DIFF
--- a/github_changelog_generator.gemspec
+++ b/github_changelog_generator.gemspec
@@ -28,4 +28,5 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency('github_api', ['~> 0.12'])
   spec.add_runtime_dependency('colorize', ['~> 0.7'])
+  spec.add_runtime_dependency('thor', ['~> 0.19'])
 end
--- a/Gemfile
+++ b/Gemfile
@@ -1,9 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'rake'
-
-gem 'github_api'
-gem 'colorize'
+gemspec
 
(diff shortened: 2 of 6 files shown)


skywinder:
It looks very neat and self-descriptive, unlike OptionParser. So, I think it's a good idea to go with `thor`! :+1:

tya:
+1 for this change.  This auto version bump tool also uses thor: https://github.com/RiotGamesMinions/thor-scmversion

skywinder:
@tya Ok. I will try to play around this versionIgniting system as well. Thanks you for suggestion!
I don't forget about this PR, @estahn take a great work to improve this project and I so appreciate that! As soon as I complete my currents tasks - I will handle this PR.

github_changelog_generator.gemspec
```diff
@@ -28,4 +28,5 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency('github_api', ['~> 0.12'])
   spec.add_runtime_dependency('colorize', ['~> 0.7'])
+  spec.add_runtime_dependency('thor', ['~> 0.19'])
 end
```
martymcfly23: /> <H3> ###

skywinder: @martymcfly23 what does it mean?

skywinder:
Thanks you, @estahn 
Thor is incredibly powerful feature, but seems it's not time for that merge. 👍

PR #231, opened 2015-05-06
Add future release option

### LINKED ISSUE #230: Trees/Archives with missing change log notes for the current tag.
Hello @skywinder !
very nice changelog generator :heart_eyes:
Maybe im not understanding something, but to me looks like something is *wrong*.

Using your lovely project as example, if you go to tree 1.3.9, in its changelog, it only appears notes until 1.3.8, and you need to go to 1.3.10 to be able to read the notes of 1.3.9.

See change log file of tree 1.3.9 with missing change log notes for 1.3.9 at https://github.com/skywinder/github-changelog-generator/blob/1.3.9/CHANGELOG.md

The real *wrong* thing here for my eyes, is that the archives for 1.3.9 misses the change log notes for 1.3.9 as seen at https://github.com/skywinder/github-changelog-generator/archive/1.3.9.tar.gz

Obviously, this occurs because the generation of the change log file happens to be after the creation of the tag, and therefore, the commit of the change log notes happens to fall into the next tag.

But sadly, it stills looks to me that each tree/archive should contain its own change log notes.

Did you thought about it?

TL:DR: aww very niiicee change log generator! can we include change log notes of the current tag in their respective archives/trees *somehow*?

sildur:
Adds an option to put all unreleased changes into a future tag specified by parameter. Works like unreleased-label, but it also modifies the links to point to a [yet] inexistent release, soon to be created.
This fixes #230.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -7,7 +7,7 @@
 
 # Offense count: 21
 Metrics/AbcSize:
-  Max: 71
+  Enabled: false
 
 # Offense count: 2
 Metrics/BlockNesting:
@@ -25,7 +25,7 @@ Metrics/CyclomaticComplexity:
 # Offense count: 28
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
-  Max: 118
+  Enabled: false
 
 # Offense count: 5
 Metrics/PerceivedComplexity:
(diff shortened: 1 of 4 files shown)


skywinder:
Great addition! 
Please update readme with this new option and fix a couple of rubocop warnings to not fail builds in Travis (https://travis-ci.org/skywinder/github-changelog-generator/builds/61473930)

alemata:
@skywinder we fixed a few rubocop warnings by disabling some checks, because we feel that there's no point to just increasing the check parameter when the check fails. On the other hand coverage decreased slightly because code was added and there are few tests

skywinder:
@alemata fully agree with your changes in rubocops checks. Thanks for update! :+1:

PR #238, opened 2015-05-18
Wrap GitHub requests in function check_github_response

skywinder:
All `begin-rescue-end` blocks for  `@github` parameter in `fetcher.rb`, that check that github fetch completed without errors. moved to (`check_github_response` function). 
Not I have to call `check_github_response` before all `@github` calls.

### DIFF
--- a/lib/github_changelog_generator/fetcher.rb
+++ b/lib/github_changelog_generator/fetcher.rb
@@ -9,8 +9,10 @@ module GitHubChangelogGenerator
   class Fetcher
     PER_PAGE_NUMBER = 30
     CHANGELOG_GITHUB_TOKEN = "CHANGELOG_GITHUB_TOKEN"
-    GH_RATE_LIMIT_EXCEEDED_MSG = "Warning: GitHub API rate limit (5000 per hour) exceeded, change log may be " \
-        "missing some issues. You can limit the number of issues fetched using the `--max-issues NUM` argument."
+    GH_RATE_LIMIT_EXCEEDED_MSG = "Warning: Can't finish operation: GitHub API rate limit exceeded, change log may be " \
+    "missing some issues. You can limit the number of issues fetched using the `--max-issues NUM` argument."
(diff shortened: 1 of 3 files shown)


skywinder:
@estahn since you are my Sensei in ruby development - I would ask you a advice:
I have a lot of similar begin-rescue-end blocks for @github parameter in fetcher.rb, that check that github fetch completed without errors. I decide to place all this checks in one place (check_github_response function). before perform this to all others github functions - I would like to hear your feedback: is this a good practice to wrap it with yield in another function? Or there is some more easier way to wrap all actions with parameter @github in fetcher.rb with begin-rescue-end block?

lib/github_changelog_generator/fetcher.rb
```diff
@@ -29,11 +31,7 @@ def initialize(options = {})
       github_options[:endpoint] = options[:github_endpoint] unless options[:github_endpoint].nil?
       github_options[:site] = options[:github_endpoint] unless options[:github_site].nil?
 
-      begin
-        @github = Github.new github_options
-      rescue
-        @logger.warn GH_RATE_LIMIT_EXCEEDED_MSG.yellow
-      end
+      @github = check_github_response { Github.new github_options }
```
skywinder: @estahn I'm talking about this function

lib/github_changelog_generator/fetcher.rb
```diff
@@ -60,29 +57,41 @@ def get_all_tags
...
-        response = @github.repos.tags @options[:user], @options[:project]
-        page_i = 0
-        count_pages = response.count_pages
-        response.each_page do |page|
-          page_i += PER_PAGE_NUMBER
-          print "Fetching tags... #{page_i}/#{count_pages * PER_PAGE_NUMBER}\r"
-          tags.concat(page)
-        end
-        print "                               \r"
+      check_github_response { github_fetch_tags(tags) }
```
skywinder: and this

skywinder:
@estahn I merge this PR, but not sure, that is correct way to do such things, so I'm waiting for your response before apply it to all other functions.

PR #247, opened 2015-06-11
Fix the tag finding

ErjanGavalji:


skywinder:
Thanks for PR, But I already implement it and fix this bug (and tests also). I will release new version today

skywinder:
Fixed in 6fdcc0c

skywinder:
@ErjanGavalji anyway, much appreciated. Thanks! :+1:

ErjanGavalji:
! perfect !

PR #257, opened 2015-08-04
Add --since-tag, close 

### LINKED ISSUE #254: Generate change log since/due specific tag
Hi, is there a way to generate the log since a particular tag? The `--between-tags` option doesn't seem to allow this. I tried doing a hack where I used git to get the latest tag and then pass it to the `github_changelog_generator` command, but the output is different from what I expected. The `--between-tags` option seems to roll up all the pull requests in between into the later tag, instead of listing each tag in between. I did not expect that, but I assume that is the intended behavior?

raphink:
Fix #254

### DIFF
--- a/lib/github_changelog_generator/generator/generator_tags.rb
+++ b/lib/github_changelog_generator/generator/generator_tags.rb
@@ -56,8 +56,27 @@ module GitHubChangelogGenerator
     #
     # @return [Array]
     def get_filtered_tags(all_tags)
-      filtered_tags = filter_between_tags(all_tags)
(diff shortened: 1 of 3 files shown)


skywinder:
Hi, @raphink! Awesome addition!
Could you please add  basic tests for new method? It's should be quite similar to `filter_between_tags` method test: https://github.com/skywinder/github-changelog-generator/blob/master/spec/unit/generator/generator_tags_spec.rb#L17-L62
Then I'll be happy to merge it!
Related with #254.

PR #261, opened 2015-08-21
Fixed Readme, I think

dpwspoon:
I don't know where a 16 digit token comes from.  But the 40 digit token works.  So I think this is a typo.


dpwspoon:
Not sure why the huge diff. Github generated the pull request for me and all I changed was the "16 digit" line.   maybe something to do with spaces

skywinder:
Yep, you're right. Thanks!

PR #260, opened 2015-08-05
Add a rake task

raphink:

PR #264, opened 2015-08-24
Add release_url to rake task options

raphink:
This made it through my previous PRs as it required both #259 and #260.


skywinder:
:hammer:

PR #267, opened 2015-08-25
Use since_tag as default for older_tag

raphink:
When using `since_tag`, the full changelog link is not generated for the first entry.

This commit makes `older_tag` default to `since_tag` when `nil` to fix this.


skywinder:
It will fix problem, described here: https://github.com/skywinder/github-changelog-generator/pull/258#issuecomment-134245090

raphink:
It might yes, but it won't fix the fact that the main header (`# Change Log`) will be repeated imo.

PR #278, opened 2015-09-14
Auto parse options from file .github_changelog_generator

dlanileonardo:
In the project you can put options file example:

.github_changelog_generator
```
future-release=5.0.0
since-tag=1.0.0
```

And the gem add to options instance.


skywinder:
@dlanileonardo Hi! Thanks, I really love this way of storage setting for specific repos. :+1: :tada: 

I have a several requests for code improvement before merge:

1. The priority of options, that user put in command line should be higher, than priority of options, specified in this file. So, lets parse `.github_changelog_generator` file first, and then parse options and override is some options specified in command line.

2. Please, add node in `README.md` about this option.

3. Since project getting bigger, so I start to cover all new features with tests - please add basic tests for your functionality also. (at least 4 tests: empty file, correct file, incorrect file, overriding of default values)

dlanileonardo:
Ok! I will follow your requests and commit in soon.

dlanileonardo:
Done! You can check?

skywinder:
@dlanileonardo Awesome work! Thanks! :beers: 
Almost ready to merge, just a couple of notices

.gitignore
```diff
@@ -0,0 +1 @@
+*.swp
```
skywinder: Since it's not related to project, I believe it should be in `.gitignore_global` file, not in local repo.

Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    github_changelog_generator (1.8.1)
+    github_changelog_generator (1.9.0)
```
skywinder: I will update version with release (it will bumped automatically)
please, revert this bump

lib/github_changelog_generator/parser_file.rb
```diff
@@ -0,0 +1,36 @@
...
+          value = false if value =~ (/^(false|f|no|n|0)$/i)
+          @options[key_sym] = value
+        rescue
+          raise "File #{file} is incorrect in line \"#{line.gsub(/[\n\r]+/, '')}\""
+        end
+      end
+      @options
+    end
+  end
+end
```
skywinder: Separate file for it. Good idea, thanks! :+1:

lib/github_changelog_generator/version.rb
```diff
@@ -1,3 +1,3 @@
 module GitHubChangelogGenerator
-  VERSION = "1.8.5"
+  VERSION = "1.9.0"
```
skywinder: and  this string also

lib/github_changelog_generator/parser_file.rb
```diff
@@ -0,0 +1,36 @@
...
+    def has_file?
+      File.exist?(file)
+    end
+
+    def file_open
+      File.open(file)
+    end
+
+    def parse!
+      return false unless has_file?
```
skywinder: Why `false`, not  just `nil`?
I believe it's better to return only one type (`Hash` in this case) from method or `nil`. It's MHO. I didn't find any guidelines about that.  Thoughts?

skywinder:
And last one, just to pass https://travis-ci.org/skywinder/github-changelog-generator/builds/80491796 you have to rub `rubocop -a` to autocorrect some code-style issues. it's not mandatory, just for green circle od check before merge :wink:

dlanileonardo:
Hi! Done, are you think? Its ok? :smile:

skywinder:
@dlanileonardo It's simply Great! :tada:

PR #279, opened 2015-09-14
Rake task: Be able to set false value in config

olleolleolle:
I wanted to disable the author information from the changelog, and noted that I couldn't. 

Instead of checking for falsy config values from the Rake task's configuration block, let's check-and-skip  nil values.

Example of what works, now:

```ruby
GitHubChangelogGenerator::RakeTask.new(:changelog) do |config|
  config.author = false
end
```


skywinder:
@olleolleolle Thanks! :+1:

PR #281, opened 2015-09-15
Feature: exclude_tags using regular expression

olleolleolle:
This PR now allows the user to filter out tags from the changelog **using regular expression patterns**.

```ruby
GitHubChangelogGenerator::RakeTask.new do |options|
  options.exclude_tags = /2013\.\d+\.\d+/
end
```

Question: Would it be reasonable only to support this specific filtering in the Ruby-based options? Or do we need to invent a way to allow users to configure this from command-line, too?


skywinder:
@olleolleolle Great addition and very neat code! Thanks! :+1:

About question, I'm not clearly understand it, what you mean under "Ruby-based options"?

olleolleolle:
Ruby-based options: it is easy to express that something _is_ a Regexp in a block in a Rake task.

When trying to use this **from the command-line options**, it may be harder. 

Adding an extra step, preparing this options: One way to figure out that given `--exclude_tags` is not a list of tagnames, but a pattern, would be to ask "Is the first character a forward-slash? Then it is a Regexp."

PR #297, opened 2015-10-21
Rubocop: less complex methods in parser.rb

olleolleolle:
- Changed one thing: failed return value on printing the banner when
    no user or project was found.
- Renamed a `get_` method to fix a Rubocop TODO


skywinder:
Looks very neat! Thanks, @olleolleolle! :+1:

PR #299, opened 2015-10-23
[refactor] Fix docblock datatype, use #map

olleolleolle:
The text is delimited into sections, so each output element from `parse` could be called a Section. So I renamed a local variable.

And, `#parse` collects all the Sections and returns them.

And the doc block is fixed, too.

### DIFF
--- a/lib/github_changelog_generator/reader.rb
+++ b/lib/github_changelog_generator/reader.rb
@@ -62,22 +62,19 @@ module GitHubChangelogGenerator
       captures
     end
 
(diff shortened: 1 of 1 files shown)


skywinder:
Hey, you firmly took the project in your hands! :+1: :wink:

skywinder:
Sorry, that can't help you now, have a lot urgent tasks, @olleolleolle.

olleolleolle:
Thank you for being so open and welcoming with your project.

PR #301, opened 2015-10-25
Use GitHub tags to reference GitHub users

phase:
This will use @tags to notify the contributor that their contribution is in the changelog.

As discussed in https://github.com/github/linguist/pull/2698#issuecomment-150887871

### DIFF
--- a/lib/github_changelog_generator/generator/generator_generation.rb
+++ b/lib/github_changelog_generator/generator/generator_generation.rb
@@ -172,7 +172,11 @@ module GitHubChangelogGenerator
           if issue.user.nil?
             title_with_number += " ({Null user})"
           else
-            title_with_number += " ([#{issue.user.login}](#{issue.user.html_url}))"
+            if issue.user.html_url.exclude? "github.com"
+              title_with_number += " ([#{issue.user.login}](#{issue.user.html_url}))"
+            else
+              title_with_number += " (@#{issue.user.login})"
+            end
           end
         end
       end
--- a/lib/github_changelog_generator/generator/generator_generation.rb
+++ b/lib/github_changelog_generator/generator/generator_generation.rb
@@ -168,15 +168,15 @@ module GitHubChangelogGenerator
       title_with_number = "#{encapsulated_title} [\\##{issue[:number]}](#{issue.html_url})"
       issue_line_with_user(title_with_number, issue)
     end
-    
+
     private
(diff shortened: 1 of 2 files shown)


olleolleolle:
Cool feature! Thanks for proposing it as code!

I think of it as a short-form, GitHub-only Markdown optimization.

I tried extracting user presentation to its own method https://gist.github.com/olleolleolle/218060ffe29017d7d9f1 - which avoided any Rubocoply warnings - but changed no behavior. This can then be a site for any future option-querying of settings like `@options["usernames_as_github_logins"]`.

phase:
@olleolleolle Should be good to go!

olleolleolle:
- Before this change, links to user profiles would work anywhere Markdown could be generated
- This PR introduces a slight change of behavior – favouring rendering on GitHub over "anywhere Markdown can be rendered"

If @skywinder finds the "render username as a link anywhere" functionality useful enough to keep, we may need to introduce a user-configurable option.

skywinder:
Agree with @olleolleolle about user-configurable option. 
The feature make change log a quite shorter, but we lose ability to follow by user's link anywhere but GitHub page.
I agree to add this feature as option that disabled by default (to save backward compatibility). `@options["usernames_as_github_logins"]` seems clear to understand.

phase:
Maybe something like `tag_github_users`. How would I implement this as an option?

olleolleolle:
@phase There are a few ways of specifying options

- rake options
- CLI options

Here is the OptionParser-based CLI option parsing, where you can introduce `usernames_as_github_logins`

https://github.com/skywinder/github-changelog-generator/blob/master/lib/github_changelog_generator/parser.rb#L39

olleolleolle:
And, the Rake task options are picked up here: https://github.com/skywinder/github-changelog-generator/blob/master/lib/github_changelog_generator/task.rb#L9

phase:
Something like that? (I will add a check for it in a second)

olleolleolle:
We want to inform the user of what the option does ("recommended for rendering on GitHub" vs "recommended for rendering anywhere") and what the default value is.

skywinder:
@phase are you going to finish this PR? Seems `usernames-as-github-logins` not binded to new option yet.

olleolleolle:
Closing this issue, carried in PR #301.

PR #315, opened 2016-01-05
Parser: avoid Ruby exit, to make Rake tasks work

olleolleolle:
- Change: made the `Parser.user_project_from_option`  method require three arguments

This made my Rake task, configured like this, work; 1.10.0 hit an `exit` with a confusing error message:

```ruby
  GitHubChangelogGenerator::RakeTask.new(:changelog) do |config|
    config.author = false
    config.exclude_tags = /201[^4]|a.exe/
    config.release_branch = '2014.1'
  end
```

The confusing error message was:

```
> $ bundle exec rake releases:rebuild_changelog
releases:rebuild_changelog
Can't detect user and name from first parameter: 'releases:rebuild_changelog' -> exit'
```

(What happens in ARGV in my Rake task is not the concern of this class. We should work our way away from that bit. See line 198 in parser.rb.)

PS: The linting of the gemspec made Travis happier.

### DIFF
--- a/github_changelog_generator.gemspec
+++ b/github_changelog_generator.gemspec
@@ -9,14 +9,14 @@ Gem::Specification.new do |spec|
   spec.version            = GitHubChangelogGenerator::VERSION
   spec.default_executable = "github_changelog_generator"
 
-  spec.required_ruby_version     = ">= 1.9.3"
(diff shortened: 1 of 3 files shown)


skywinder:
Already in `1.10.1` Thanks! :+1:

PR #320, opened 2016-01-22
Added tag exclusion with a filter (string or regex)

soundstep:
Hi,

Here is a pull request to be able to exclude tag name with a regular expression.
I give you our use case, we have tags created by Jenkins including a build number, it looks like this:

```
v1.0.0+1234
```

We wanted to generated a changelog for all tags expect the one with a build number (+1234).

The command would look like this:

```
github_changelog_generator --exclude-tags-regex ".*\+\d{1,}"
```

Quick note: this is the first time I write ruby, so I did what I could. I wanted to add a rspec test for it, but it is a bit beyond me, If anyone wants to help, feel free!

We can use the forked version in our pipeline, but if that PR makes sense to you, here you go!

Cheers.

### DIFF
--- a/lib/github_changelog_generator/fetcher.rb
+++ b/lib/github_changelog_generator/fetcher.rb
@@ -74,7 +74,7 @@ module GitHubChangelogGenerator
         if @options[:filter_tags]
           body_new = []
           reg = Regexp.new @options[:filter_tags]
-          page.body.each_with_index do |tag,index|
+          page.body.each do |tag|
             if !(tag.name =~ reg)
               body_new << tag
             end
--- a/lib/github_changelog_generator/fetcher.rb
+++ b/lib/github_changelog_generator/fetcher.rb
@@ -68,11 +68,25 @@ module GitHubChangelogGenerator
       response = @github.repos.tags @options[:user], @options[:project]
       page_i = 0
       count_pages = response.count_pages
+
       response.each_page do |page|
+
+        if @options[:filter_tags]
+          body_new = []
+          reg = Regexp.new @options[:filter_tags]
+          page.body.each_with_index do |tag,index|
+            if !(tag.name =~ reg)
+              body_new << tag
+            end
+          end
+          page.body = body_new
+        end
+
         page_i += PER_PAGE_NUMBER
         print_in_same_line("Fetching tags... #{page_i}/#{count_pages * PER_PAGE_NUMBER}")
         tags.concat(page)
       end
+
       print_empty_line
 
       if tags.count == 0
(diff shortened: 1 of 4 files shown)


lib/github_changelog_generator/parser.rb
```diff
@@ -116,6 +116,9 @@ def self.setup_parser(options)
         opts.on("--between-tags  x,y,z", Array, "Change log will be filled only between specified tags") do |list|
           options[:between_tags] = list
         end
+        opts.on("--exclude-tags-filter [REGEX]", "Apply a regular expression on tag names so that they can be excluded, for example: --exclude-tags-filter \".*\+\d{1,}\" ") do |last|
+          options[:exclude_tags_filter] = last
```
olleolleolle: Late feedback: why is this block parameter called `last` (where the others are called `list`)?

soundstep: Not sure, there a bunch of parameters called ```list``` and another bunch of parameters ```last```.
See there: 
https://github.com/ITV/github-changelog-generator/blob/filter-tags/lib/github_changelog_generator/parser.rb
I didn't create all these.

soundstep: I can change it to list if you want though.

lib/github_changelog_generator/fetcher.rb
```diff
@@ -68,11 +68,25 @@ def github_fetch_tags
...
       count_pages = response.count_pages
+
       response.each_page do |page|
+
+        if @options[:exclude_tags_filter]
+          body_new = []
+          reg = Regexp.new @options[:exclude_tags_filter]
+          page.body.each do |tag|
+            if !(tag.name =~ reg)
+              body_new << tag
```
skywinder: Looks nice.
Please add test to cover new feature.

testerindevelopment:
Hello @skywinder
We've added some tests and refactored a bit
The branch is ready to merge

spec/unit/generator/generator_tags_spec.rb
```diff
@@ -76,31 +76,47 @@ def tags_mash_from_strings(tags_strings)
...
+      it { is_expected.to be_a Array }
+      it { is_expected.to match_array(tags_mash_from_strings(%w(1))) }
+    end
+
+    context "with non-matching regex" do
+      let(:generator) { GitHubChangelogGenerator::Generator.new(exclude_tags_regex: '[45]') }
+      it { is_expected.to be_a Array }
+      it { is_expected.to match_array(tags_mash_from_strings(%w(1 2 3))) }
+    end
+  end
```
olleolleolle: Nice use of the oneliner syntax and `let`s!

skywinder:
@testerindevelopment @soundstep You rocks, guys, thank you!

PR #321, opened 2016-02-03
Four new features

retorquere:
Four new features:
- HTTP response caching to prevent rate limit errors
- Filter empty tags
- Exclude tags by regex
- ~~add YAML [front-matter](http://jekyllrb.com/docs/frontmatter/) so the generated changelog can be put on GH pages directly~~


github_changelog_generator.gemspec
```diff
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency("github_api", ["~> 0.12"])
   spec.add_runtime_dependency("colorize", ["~> 0.7"])
+  spec.add_runtime_dependency("sqlite3")
```
olleolleolle: Oh! Here be jdbc dragons - perhaps we can improve it for users on JRuby?

Compare [this change](https://github.com/earaujoassis/sequel-seed/commit/9a470a97949efe41d505cc21f3bcbba8c816070d), to a completely different project, for inspiration.

lib/github_changelog_generator/generator/generator_generation.rb
```diff
@@ -8,7 +8,9 @@ def compound_changelog
       sort_tags_by_date(@filtered_tags)
       fetch_issues_and_pr
 
-      log = "#{@options[:header]}\n\n"
+      log = ''
+      log += @options[:frontmatter] if @options[:frontmatter]
```
olleolleolle: This is supernice.

github_changelog_generator.gemspec
retorquere: I'm not really familiar with JRuby, so it's not something I could test. It doesn't even have to be sqlite, I just needed something that would autocommit without too big of a performance hit (end preferably not a big heap of files).

retorquere: I could try replacing it with http://ruby-doc.org/stdlib-2.1.3/libdoc/yaml/rdoc/YAML/Store.html . I don't know the performance characteristics of YAML::Store, but it wouldn't need any gems.

lib/github_changelog_generator/parser.rb
```diff
@@ -117,7 +123,11 @@ def self.setup_parser(options)
           options[:between_tags] = list
         end
         opts.on("--exclude-tags  x,y,z", Array, "Change log will exclude specified tags") do |list|
-          options[:exclude_tags] = list
+          if list.length == 1 && list[0].length > 2 && list[0][0] == '/' && list[0][-1] == '/'
```
olleolleolle: :golf: 

- String#end_with?
- String#start_with?
- Enumerable#first

`if list.length == 1 && list.first.length > 2 && list.first.start_with?('/') && list.first.end_with?('/')`

retorquere: I've replaced it with a simple regex match.

lib/github_changelog_generator/generator/generator_generation.rb
retorquere: Sure. I was wondering about lumping them all together, but testing without the caching became incredibly frustrating.

github_changelog_generator.gemspec
olleolleolle: Cool idea! The changes would not ripple very much, I guess.

retorquere: No, the use of SQLite is very localized.

Sorry about the huge monkey-patch for caching. I couldn't find any other way to inject the cacher; the github_api ought to support it but it has been broken over a year.

lib/github_changelog_generator/generator/generator_generation.rb
retorquere: PStore (which is like YAML::Store but thread safe) is a *lot* slower than SQLite. To the point where you might not have to worry about rate limiting, because PStore access spreads the activity anyway.

github_changelog_generator.gemspec
retorquere: I just tried, but PStore (YAML::Store inherits PStore, it just changes the serializer) is unusably slow.

olleolleolle: Aw shucks.

Well, I can work with you to make the access to Sqlite platform-indepdendent.

Speed improvements for the users is quality work.

retorquere: I'm going to give it a go using https://github.com/ruby-concurrency/thread_safe, loading the entire structure into mem at start, and using an at_exit block to save. Downside is that interrupted runs might lose the entire cache, but PStore negated the entire idea of the cache.

retorquere: OK so that *sort of* works, but saving the whole cache takes a *looooong* time. It's been sitting at "Generated log placed in..." for over two minute now, as it needs to serialise and save all entries always instead of just the updated ones. Assuming that [this](https://github.com/jruby/dbm) will load when I go "require dbm" I'm willing to give that a shot, but SQLite is the only thing that's performed adequately so far.

olleolleolle: (It'll only load as jruby-dbm if we make it to.)

(Could we turn things upside-down - have a new problem - try having VCR help us do the request caching?)

retorquere: DBM is giving me permission denied errors; haven't yet been able to trace the cause, as they don't happen always.

VCR won't help since it doesn't send out Etags, it just assumes that whatever you get is valid for a period you choose beforehand.

So far, SQLite is the only thing that has worked for me both reliably and fast.

retorquere: I don't know how the line in Gemfile (in the inspiration piece you linked to for jruby sqlite) translates to gemspec.

retorquere: Would this work to load the sqlite gem?

```
begin
  require 'jdbc/sqlite3'
  Jdbc::SQLite3.load_driver
rescue
  require "sqlite3"
end
```

retorquere: I've also tried faraday-http-cache but it didn't seem to save anything to the cache even though the middleware was being called.

skywinder:
@retorquere 
Thanks, looks very promising! 
It's better to split this features to different PR. But since you get a lot of work done, let's leave it in one PR.
What the current status of this PR?

skywinder:
This PR will close #361

retorquere:
It doesn't look like it'll merge cleanly at this point. I developed these together because I needed the release squashing, but the rate limits meant I couldn't sensibly test my work. That's why they're both here; I couldn't not have the caching on my branch.

olleolleolle:
@retorquere I am sorry that this PR rotted on the vine.

Is it alright if I close it? 

Here are my immediate thoughts on what went in this change, related to what's in master, now:

- Options parsing: Regular expression from a string, this change _could_ be work to be done in #552 (I'm not closely familiar with the details, so this is me connecting dots more than making a deep investigation.)
- Error level logs - at the moment, all coloring is based on log's severity level, so it happens automatically. So, I consider that change fixed.
- HTTP response caching - if it is broken with the middleware in place, it's a defect, and should be logged as an Issue
- SQLite dumps of each call. We're now _using_ Faraday and its middleware, so to make a change like this, we could be using custom middlewares.

retorquere:
No problems to close this.

PR #322, opened 2016-02-03
YAML front matter

retorquere:


olleolleolle:
Gosh, a Rubocop complaint. 

lib/github_changelog_generator/generator/generator_generation.rb:11:13: C: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

Other than that, it looks merge-ready to me.

Issue #327, opened 2016-02-22
Error when specifying exclude_labels


joffrey-bion:
When I specify the `exclude_labels` property in the `.github_changelog_generator` file, I get the following error:

    generator_processor.rb:9:in `&': no implicit conversion of String into Array (TypeError)
    from generator_processor.rb:9:in `block in exclude_issues_by_labels'
    from generator/generator_processor.rb:8:in `select'
    ...

I tried different syntaxes in the config file, but none work:

    exclude-labels=duplicate,on hold,wontfix
    exclude-labels=[duplicate,on hold,wontfix]
    exclude-labels=["duplicate","on hold","wontfix"]

This does not occur when I specify it on the command line:

    github_changelog_generator --exclude-labels=wontfix,duplicate

jlambert121:
I'm having a slightly different error installing, both from rubygems and master on GH.  This is on OSX 10.11.4 (beta), ruby 2.0.0p648, gem 2.0.14.1.

```
Building native extensions.  This could take a while...
ERROR:  Error installing github_changelog_generator:
	ERROR: Failed to build gem native extension.

    "/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby" -rubygems /Library/Ruby/Gems/2.0.0/gems/rake-10.4.2/bin/rake RUBYARCHDIR=/Library/Ruby/Gems/2.0.0/gems/github_changelog_generator-1.11.0/lib RUBYLIBDIR=/Library/Ruby/Gems/2.0.0/gems/github_changelog_generator-1.11.0/lib
fatal: Not a git repository (or any of the parent directories): .git
rake aborted!
LoadError: cannot load such file -- rubocop/rake_task
/Library/Ruby/Gems/2.0.0/gems/github_changelog_generator-1.11.0/Rakefile:2:in `<top (required)>'
(See full trace by running task with --trace)


Gem files will remain installed in /Library/Ruby/Gems/2.0.0/gems/github_changelog_generator-1.11.0 for inspection.
Results logged to /Library/Ruby/Gems/2.0.0/gems/github_changelog_generator-1.11.0/./gem_make.out
```

skywinder:
@jlambert121 try again. It should work now.

olleolleolle:
Continued work on this in #354.

PR #331, opened 2016-02-23
Test case for #327

olleolleolle:
Here is a failing test-case for part of #327 - the "read options from a file" code does not apply the OptionParser type conversions.


skywinder:
that's how unit test should works. Thanks @olleolleolle for your perfect code. I learn from you best coding practice with every commit! (actually I'm not ruby developer, my specialty is iOS, but relly like ruby. So I very glad that you help me that way!  You are the best! :dizzy:

olleolleolle:
@skywinder Glad that someone reads the code!

Now, we only need to fix this failing test.

(Weird trick in this commit: When test data is meaningless, I make it uniquely findable. `brew install ossp-uuid` in order to have CLI access to a UUID generator.)

PR #332, opened 2016-02-23
[Refactor] Generator#exclude_issues_by_labels simpler, tested

olleolleolle:
This PR wants to bring attention to the `&` operator being used for its filtering.


skywinder:
much cleaner code, thanks again!

PR #337, opened 2016-02-25
Fix installation by not running the specs - which have dependencies

skywinder:
Remove check from default rake rask


olleolleolle:
:+1: r=@olleolleolle

skywinder:
@olleolleolle Ok! And Travis still runs all checks. That's good! So, seems everything is ok with this fix. Thanks for participation!

PR #343, opened 2016-02-29
Gemspec: Calculate date using Date stdlib

olleolleolle:
http://ruby-doc.org/stdlib-2.3.0/libdoc/date/rdoc/Date.html#method-c-today


skywinder:
[According this](http://ruby-doc.org/stdlib-1.9.3/libdoc/rubygems/rdoc/Gem/Specification.html#method-i-date) - yes, it's not required at all.

PR #353, opened 2016-03-09
Add a `bundle install` test

jkeiser:
I am mainly doing this to find out if Travis is broked


.gitignore
```diff
@@ -1 +1,7 @@
+bin/
+!bin/git-generate-changelog
+!bin/github_changelog_generator
```
olleolleolle: Should these really go in here?

olleolleolle:
I like this test - and I had a question about one of the gitignores.

jkeiser:
Anyone who uses bundle install --binstubs will put lots of binaries in there. Only reason for the gitignore

olleolleolle:
Thanks, grand! :+1:

Issue #312, opened 2015-12-22
.github_changelog_generator config file is not consistent with the internal options hash


stanislav-zaprudskiy:
Currently, an option expected from the user is not usually consistent with the corresponding option hash key. For instance, `--enhancement-label [LABEL]` is stored as `options[:enhancement_prefix]` (`_label` part is stored as `_prefix` internally). This produces incompatibility with the `parser_file.rb` code.

That is why, there appear a situations similar to the following. One tries to use GitHub Enterprise, and provides the following entries in his `.github_changelog_generator`:
```
github-api=https://github.mycompany.com/api/v3
github-site=https://github.mycompany.com
```
Then, while invoking the binary, he receives the following error:
```
Bad credentials.
Error: wrong GitHub token
```
Since `parser_file.rb` placed `github-api` configuration entry as `:github_api` key in the hash, while `:github_endpoint` was expected. Then `Github.new` is being invoked as
```
{:per_page=>30, :oauth_token=>"eb188ed0b7ee67a7cf714f021ef2fcf48177ba75", :site=>nil}
```
which actually causes the error above. And that was not very straightforward to troubleshoot.

Do you think that is wise to add some additional conditionals/logics into `parser_file.rb` to deal with such non-conformant options? I might be able to take this on.

/CC: @dlanileonardo

olleolleolle:
This is now released.

Issue #316, opened 2016-01-08
Add support for custom sections


SandroMachado:
Currently the project groups the `merged` pull-requests, the issues with the label `bug` and the issues with the label `enhancement`. All the issues with other labels are grouped at the same block. 

Is it possible to add more labels to group the issues with that labels?  

Thanks for your work.

skywinder:
No, there is no such option yet.

ruipenso:
:+1:

olleolleolle:
(In order to know more about the feature, here's a guess about how it could be used.)

> Randy the Rake-User wants to group cool new stuff under a release notes Section named "Goodies in this release", which lists any `merged` pull-request which has the label `goodies` or the label `wonderful`.

> He introduces this new section by amending his Rake task definition:

```ruby
config.custom_sections << ['Goodies in this release', %w(goodies wonderful)]
```

skywinder:
@olleolleolle :+1: 

or by command line smth like:
`--custom_sections 'Goodies in this release', goodies, wonderful`

But how to make ability to add several custom sections? 
Obviously, we have to create internal Dictionary `[String:Array]`
like 
```
{
"Bugs" : [bug, Bug, another Bug Label],
"Features" : [feature, improvement],
...
}

```

Do you know convenient way to add it via CLI options?

ruipenso:
@skywinder Any news? :)

skywinder:
@ruipenso didn't work on that feature yet.

aerosol:
Just came here to give big ups :100:. Being able to group the changes per custom labels, e.g. "API" and "Front-End" would be totally awesome.

The CLI interface could look like:

```
--custom-group-{name}-labels=foo,bar,baz
```

e.g.

```
--custom-group-api-labels=api --custom-group-frontend-labels=web,javascript
```

This way you could easily (sorry if I'm underestimating this - I'm not fluent with ruby at all), build the hierarchy - first build the custom groups, then within those, build the log of bugs, enhancements etc.

skywinder:
@aerosol how to solve several groups in that case?

aerosol:
@skywinder could you give an example?

nicwaller:
I would also appreciate this feature! In accordance with [Semantic Versioning](http://semver.org/) I make a point of highlighting changes that break backward compatibility. It would be great if I could configure the changelog generator to include a section for breaking changes. Like this.

```
[...]

PR #354, opened 2016-03-17
Parse options file options into arrays, integers, flags, and other

### LINKED ISSUE #327: Error when specifying exclude_labels

olleolleolle:
The file-based `.github_changelog_generator` options support has been less than perfect. This PR wants to do the same options translation that happens in the OptionParser in the `parser.rb` file, for the file-based settings.

- Respect each option's type (Array, Integer, flag, anything), according to #327 
- Rename each given setting to its option, according to #312.

- Q: Do we need better parsing of the array values than `#split(",")`? Answer: [no, this is what OptionParser's Array coercion does](http://ruby-doc.org/stdlib-2.3.0/libdoc/optparse/rdoc/OptionParser.html#class-OptionParser-label-Type+Coercion)

Fix #327

### DIFF
--- a/lib/github_changelog_generator/parser_file.rb
+++ b/lib/github_changelog_generator/parser_file.rb
@@ -23,9 +23,7 @@ module GitHubChangelogGenerator
 
     def parse_line!(line)
       key_sym, value = extract_pair(line)
-      value = true if value =~ /^(true|t|yes|y|1)$/i
-      value = false if value =~ /^(false|f|no|n|0)$/i
-      @options[key_sym] = value
+      @options[key_sym] = convert_value(value, key_sym)
     rescue
       raise ParserError, "Config file #{file} is incorrect in line \"#{line.gsub(/[\n\r]+/, '')}\""
     end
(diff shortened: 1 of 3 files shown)


skywinder:
Very nice refactoring, @olleolleolle! :gem: 
Do you know, is there any way to add support of dictionaries also? It will help to implement #316

olleolleolle:
To answer your question:

- For Parser: OptionParser has built-in coercions but allows for other setting modes as well.
- For ParserFile: we can probably do the same in `#convert_value`.

olleolleolle:
I will merge this, as-is, and move to make other features possible in another PR.

PR #355, opened 2016-03-19
Refactor: call it option_name, instead of key_sym

olleolleolle:
This an internal name-change refactoring to ParserFile:

- call each settings key an `option_name`, instead of `key_sym`
- also, add an IRREGULAR_OPTIONS hash to explain why we translate option name to an option key
- rename a method to `#option_key_for` to avoid mixing up with i18n ("translate" can be a loaded word)


olleolleolle:
@skywinder I felt this was a missing piece from the feature addition.

Issue #349, opened 2016-03-04
Problems installing 1.11.7 on Windows when git absent


Arcanemagus:
When attempting to update to v1.11.7 on Windows I get the following error:

```ShellSession
PS C:\Users\Landon Abney> gem update
Updating installed gems
Updating github_changelog_generator
Temporarily enhancing PATH to include DevKit...
Building native extensions.  This could take a while...
ERROR:  Error installing github_changelog_generator:
        ERROR: Failed to build gem native extension.

    C:/Ruby22-x64/bin/ruby.exe -rubygems C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/rake-10.5.0/bin/rake RUBYARCHDIR=C:/Ruby22-x64/lib/ruby/gems/2.2.0/extensions/x64-mingw32/2.2.0/github_changelog_generator-1.11.7 RUBYLIBDIR=C:/Ruby22-x64/lib/ruby/gems/2.2.0/extensions/x64-mingw32/2.2.0/github_changelog_generator-1.11.7
rake aborted!
Bundler::GemspecError: There was a Errno::ENOENT while loading github_changelog_generator.gemspec:
No such file or directory - git ls-files -z from
  C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/github_changelog_generator-1.11.7/github_changelog_generator.gemspec:21:in ``'

Errno::ENOENT: No such file or directory - git ls-files -z

(See full trace by running task with --trace)

rake failed, exit code 1

Gem files will remain installed in C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/github_changelog_generator-1.11.7 for inspection.
Results logged to C:/Ruby22-x64/lib/ruby/gems/2.2.0/extensions/x64-mingw32/2.2.0/github_changelog_generator-1.11.7/gem_make.out
Gems updated: github_changelog_generator
```

It looks like this is assuming the `git` command is available when installing, which is a bit odd as no other gem has required that so far...

Arcanemagus:
Tried again from a shell that had `git` available, that is indeed the problem.

olleolleolle:
One can say the issue is that the gem wants to install its extra man pages using the Rakefile. 

You can see it being called the extension. 

If we'd skip that bit in the gemspec, the extension bit, which names the Rakefile, on your platform - it'd be able to install, I believe.

olleolleolle:
This has been changed to not use the git binary.

PR #356, opened 2016-03-19
On gem install, do not try to copy manpage files in the "extensions" step

olleolleolle:
To dodge all issues with "you need git, you need Bundler, you need ..." when performing the "extensions" step in the install process, this change wants to drop that step.

Problems to solve:
- see #349 

What we miss out on with this change: man pages will no longer be copied to a location guessed by the Rakefile. (These man pages are not automatically updated.)


olleolleolle:
I was thinking of you, @jkeiser, when I made this change.

Several small things _could_ be changed back to "development dependency" status if this were merged, right?

olleolleolle:
The dependencies I refer to can be found in the commits @jkeiser added to help us out. If we opt not to run the Rakefile as part of the install, it's all simpler.

Here, they can be found in the history of the .gemspec: https://github.com/skywinder/github-changelog-generator/commits/master/github_changelog_generator.gemspec

PR #357, opened 2016-03-19
ParserFile: fail parsing with config file line number; use a File instead of a filename

olleolleolle:
This change allows tests to pass in a file-like object instead of having to manage real files. [I had fun reading this while making this happen](https://robots.thoughtbot.com/io-in-ruby).

- drops spec fixture files for `.github_changelog_generator`, using StringIO objects next to the tests instead
- report configuration file parse failure with the line number in the error message
- docs
- improved naming


olleolleolle:
@skywinder Hi, I polished a thing.

PR #360, opened 2016-03-21
Replace shelling-out-to-Git w/ Dir call

olleolleolle:
When the user's PATH does not include `git`, the gemspec becomes unusable.

Using Ruby instead avoids this problem.

See #349


chainoy:
Fix

Issue #351, opened 2016-03-06
github_changelog_generator/generator/generator_tags.rb:61:in `detect_since_tag': undefined method `[]' for nil:NilClass (NoMethodError)


crucialfelix:
I have no idea what its trying to tell me. Should this be throwing exceptions like this ?

```
Performing task with options:
{:tag1=>nil,
 :tag2=>nil,
 :date_format=>"%Y-%m-%d",
 :output=>"CHANGELOG.md",
 :base=>"HISTORY.md",
 :issues=>true,
 :add_issues_wo_labels=>true,
 :add_pr_wo_labels=>true,
 :pulls=>true,
 :filter_issues_by_milestone=>true,
 :author=>true,
 :unreleased=>true,
 :unreleased_label=>"Unreleased",
 :compare_link=>true,
 :enhancement_labels=>["enhancement", "Enhancement"],
 :bug_labels=>["bug", "Bug"],
 :exclude_labels=>
  ["duplicate",
   "question",
   "invalid",
   "wontfix",
   "Duplicate",
   "Question",
   "Invalid",
   "Wontfix"],
 :max_issues=>nil,
 :simple_list=>false,
 :verbose=>true,
 :header=>"# Change Log",
 :merge_prefix=>"**Merged pull requests:**",
 :issue_prefix=>"**Closed issues:**",
 :bug_prefix=>"**Fixed bugs:**",
 :enhancement_prefix=>"**Enhancements**",
 :git_remote=>"origin",
 :user=>"supercollider",
 :project=>"supercollider",
 :future_release=>"3.7.0",
 :exclude_tags=>
[...]

PR #368, opened 2016-04-16
Avoid nil bug in detect_since_tag

olleolleolle:
Named a new function, moved code around in the specs (and finding shared behaviour in the specs).

See #351.


skywinder:
@olleolleolle  I'm out of PC feel a long time (may be 1 month).
Feel free to merge it, I trust you. 👍

olleolleolle:
The master branch just needed two lints, so I added a single  commit for that.

PR #379, opened 2016-05-13
Caching GH requests (but honestly, a major restructuring)

retorquere:
I kept getting lost in GCG, and since I was using sqlite for the cache anyhow, I decided to also use it as an internal data model. Not everything is done, and not everything is clean, but I think I have been able to re-implement most of what GCG does in about 300 lines of code, and (at last to me), it's a lot clearer what's going on everywhere.

This pull request is not intended to merge, but as a discussion starter to see if GCG is receptive to the changes I'd introduce.

### DIFF
--- a/bin/github_changelog_generator
+++ b/bin/github_changelog_generator
@@ -1,4 +1,301 @@
-#! /usr/bin/env ruby
+#!/usr/bin/env ruby
 
-require_relative "../lib/github_changelog_generator"
-GitHubChangelogGenerator::ChangelogGenerator.new.run
+require 'faraday/http_cache'
+require 'sqlite3'
+require 'yaml'
+require 'octokit'
+#require 'shell-spinner'
+
+require_relative "../lib/github_changelog_generator/parser"
+
+module GitHubChangelogGenerator
+  class SQLiteStore
+    @@db = nil
+    @@options = GitHubChangelogGenerator::Parser.options
+    def self.db
+      if @@db.nil?
+        @@db = SQLite3::Database.new(@@options.cache)
+        @@db.execute("CREATE TABLE IF NOT EXISTS httpcache (k PRIMARY KEY, v)")
+        @@db.execute('PRAGMA synchronous = OFF')
+        @@db.execute('PRAGMA journal_mode = MEMORY')
+        @@db.results_as_hash = true
+      end
+
+      return@@db
+    end
+
+    def read(key)
+      v = @@db.get_first_value("SELECT v FROM httpcache WHERE k = ?", key)
+      v = YAML::load(v) if v
+      print(v ? '+' : '-') if @@options.verbose
+      return v
+    end
+
+    def delete(key)
+      @@db.execute("DELETE FROM httpcache WHERE k = ?", key)
+    end
+
+    def write(key, value)
+      @@db.execute('INSERT OR REPLACE INTO httpcache (k, v) VALUES (?, ?)', [key, YAML::dump(value)])
+    end
+  end
+end
+
+DB = GitHubChangelogGenerator::SQLiteStore.db
+OPTIONS = GitHubChangelogGenerator::Parser.options
+
+Octokit.configure{|c|
+  c.access_token = ENV['CHANGELOG_GITHUB_TOKEN'] || ENV[OPTIONS.token] || OPTIONS.token
+}
+Octokit.middleware = Faraday::RackBuilder.new {|builder|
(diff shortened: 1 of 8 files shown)


github_changelog_generator.gemspec
```diff
@@ -23,12 +23,14 @@ Gem::Specification.new do |spec|
...
+  spec.add_runtime_dependency "overcommit", ">= 0.31"
   spec.add_runtime_dependency "rake", ">= 10.0"
-  spec.add_runtime_dependency("github_api", ["~> 0.12"])
-  spec.add_runtime_dependency("colorize", ["~> 0.7"])
-
-  spec.add_development_dependency("overcommit", ">= 0.31")
-  spec.add_development_dependency("rspec", ">= 3.2")
-  spec.add_development_dependency "bundler", ">= 1.7"
-  spec.add_development_dependency("rubocop", ">= 0.31")
+  spec.add_runtime_dependency "rspec", ">= 3.2"
```
skywinder: We should get rid of dev tools in runtime depedency

lib/github_changelog_generator/parser.rb
```diff
@@ -1,165 +1,181 @@
...
+        opts.on('--[no-]verbose', 'Run verbosely. Default is true') do |v|
+          @@options[:verbose] = v
         end
-        opts.on("-v", "--version", "Print version number") do |_v|
+        opts.on('-v', '--version', 'Print version number') do |_v|
           puts "Version: #{GitHubChangelogGenerator::VERSION}"
           exit
         end
-        opts.on("-h", "--help", "Displays Help") do
+        opts.on('-h', '--help', 'Displays Help') do
```
skywinder: Please, don't change double quotes to single. We would like to follow this guideline

github_changelog_generator.gemspec
retorquere: I didn't put them there, they were there when I started.

lib/github_changelog_generator/parser.rb
retorquere: Alright, pushed a change. To each their own. In any case, this push request wasn't meant to be merged, it was more a discussion starter. For one, I think most tests would fail, but I haven't yet figured out how to run the tests.

bin/github_changelog_generator
```diff
@@ -1,4 +1,301 @@
...
+end
+
+DB = GitHubChangelogGenerator::SQLiteStore.db
+OPTIONS = GitHubChangelogGenerator::Parser.options
+
+Octokit.configure{|c|
+  c.access_token = ENV['CHANGELOG_GITHUB_TOKEN'] || ENV[OPTIONS.token] || OPTIONS.token
+}
+Octokit.middleware = Faraday::RackBuilder.new {|builder|
+  builder.use Faraday::HttpCache, shared_cache: false, serializer: Marshal, store: GitHubChangelogGenerator::SQLiteStore.new
```
olleolleolle: Oh, yes, and here you _did_ use a customized middleware, my apologies.

(Reference to my bullet point item in #321.)

olleolleolle:
@retorquere So, this change is also big, and so on.

Do you wish to change the PR to fit current master?

My earlier concerns about compatibility will increase if we keep `sqlite3` as the only option; locking out JRuby users, and perhaps perhaps increasing the trouble for people on stock Ubuntu etc. (People who care about running the tool, not about maintaining their Ruby installation.)

Saving bytes transferred is the green choice, so I'm in favor of a caching layer which does what it says on the tin.

If you want, we can close this PR, and open a new Issue to be on a longer roadmap, which is about supporting a functioning caching middleware for OctoKit?

retorquere:
All good for me. I'm not specifically bound to sqlite, but I tried a heap of other persistence methods and they all had problems.

lib/github_changelog_generator/parser.rb
```diff
@@ -1,159 +1,175 @@
...
-          options[:future_release] = future_release
+          @@options[:future_release] = future_release
         end
         opts.on("--release-branch [RELEASE-BRANCH]", "Limit pull requests to the release branch, such as master or release") do |release_branch|
-          options[:release_branch] = release_branch
+          @@options[:release_branch] = release_branch
         end
         opts.on("--[no-]verbose", "Run verbosely. Default is true") do |v|
-          options[:verbose] = v
+          @@options[:verbose] = v
```
skywinder: Why do you decide to change all these variables to the class variable?

retorquere:
I have no idea. It's been a long while since I looked at this.

ferrarimarco:
Closing this, needs too much restructuring.

Issue #361, opened 2016-03-23
Cache API responses


gravitystorm:
From https://github.com/skywinder/github-changelog-generator/issues/71#issuecomment-200343511

It would be a good idea to cache the API responses (e.g. issues and PRs) to save downloading them again every time that github_changelog_generator is run.

This is a major problem for the project that I'm working on (https://github.com/gravitystorm/openstreetmap-carto) due to the moderately large number of issues and PRs (1322 and 773 so far). It is taking me a while just to implement our initial changelog, since I need to change issue tags, change config values, run github_changelog_generator again to see the results - and then I rapidly run out of API requests. In the future I will run out of API requests before the changelog can be generated just once!

Also, it takes a long time to download all the issues, so iterating our config values is painful.

I investigated the github API briefly. The list of issues (e.g. https://api.github.com/repos/gravitystorm/openstreetmap-carto/issues ) contains last-updated timestamps, so it seems possible to cache the details and only update when the timestamp changes. Do you think this would work? This way, repeated runs of github_changelog_generator would be faster and involve fewer API requests.

skywinder:
Agree, as @estahn mentioned:

> Use [octokit/octokit.rb](https://github.com/octokit/octokit.rb) with [Faraday HTTP Cache](https://github.com/octokit/octokit.rb#caching) 

could be nice alternative to #211.

retorquere:
I've started doing this on a fork

retorquere:
Not a simple change. I'm a little lost in how GCG hangs together, to be honest.

I have Octokit-based caching working, but care needs to be taken when it comes to thread-safety, and the objects Octokit yields are subtly different it seems, and the github_api objects are passed around the app, so a move to octokit touches nearly everything at once.

I'm doing some sqlite-based experiments (my current cache is sqlite -- I know about the earlier objections, but I still haven't found a good replacement), and it just struck me that GCG could probably be simplified by stashing the fetched data in sqlite, too. Filtering and matching is what databases are good at, and you get a more loose coupling internally as everything would get back hashes from the DB.

Just a thought.

skywinder:
@retorquere :+1: 
please give me name of your branch and push it in your fork to review!

To simplify this feature, I would like to finish #222 first. I want to do it this week, so it will easier to implement this feature after!

retorquere:
There's no branch yet. Nothing works right now.

mikewaters:
[...]

Issue #222, opened 2015-03-26
replace github_api to octokit


tuexss:
I think it would be good to remove the whole regex-mess and use some third-party-library like https://github.com/octokit/octokit.rb to do this in a short and clean way.

skywinder:
What kind of "regex-mess" are you meaning?

About third-party library - I agree with you. 
I get in mind to replace `github_api` to `octokit`. To work with GitHub API.
I also want to do it once:
- move all functions with API operations in separate class (like `fetcher.rb` or smth like that)
- cover it by tests
- replace `github_api` to `octokit`

Just don't have a time for it yet. :stuck_out_tongue_closed_eyes:

tuexss:
I meant the fact, that there are three different regexes and a bunch of logic to manage the three in parser.rb, when it only parses a single string in order to extract user and project.

tuexss:
Sounds like a good plan.

skywinder:
Got it. Yes, it was my first experience with regex on ruby. :smile:

mvz:
This gem now uses octokit, so I think this ticket can be closed.

olleolleolle:
Thanks all!

PR #381, opened 2016-05-18
Using Octokit for Github API requests

awaage:
This is hopefully one step in the right direction, to use Octokit for Github API requests.
This directly addresses: https://github.com/skywinder/github-changelog-generator/issues/222

I really would like to get this issue resolved: https://github.com/skywinder/github-changelog-generator/issues/361

For review purposes, I simply created a new fetcher (OctoFetcher). The follow-up to this PR should be to remove the old Fetcher. Alternatively, we could use factory pattern and support both, but I don't really see a point to it.


lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -0,0 +1,293 @@
+module GitHubChangelogGenerator
+  # A Fetcher responsible for all requests to GitHub and all basic manipulation with related data
+  # (such as filtering, validating, e.t.c)
+  #
+  # Example:
+  # fetcher = GitHubChangelogGenerator::OctoFetcher.new options
+
+  class OctoFetcher
+    PER_PAGE_NUMBER   = 100
```
awaage: Changed this to use the maximum PER_PAGE_NUMBER. This helps with large repositories, where we need to minimize API calls

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -0,0 +1,293 @@
+module GitHubChangelogGenerator
```
awaage: I tried to maintain all the interfaces, and use similar method calls, logic flow to the original Fetcher class. A follow-up PR can be done to clean it up even more, but this hopefully aids in review.

lib/github_changelog_generator/generator/generator_fetcher.rb
```diff
@@ -72,7 +72,8 @@ def set_date_from_event(event, issue)
       else
         begin
           commit = @fetcher.fetch_commit(event)
-          issue[:actual_date] = commit[:author][:date]
+          issue[:actual_date] = commit[:commit][:author][:date]
+          # issue[:actual_date] = commit[:author][:date]
         rescue
```
awaage: One difference in the way the commits are returned with Octokit. This was a nasty bug.

lib/github_changelog_generator/octo_fetcher.rb
skywinder: Perfect! 👍

skywinder:
@awaage Thank you! It's really nice work.
Can you also cover new fetcher by test at least for most important functions as well as old fetcher?

According factory - I agree with you, we don't need to handle 2 fetchers. octokit should replace it completely. 

I prefer to work on this branch separately and merge to master, when old fetcher will completely eliminated.

awaage:
I added specs for original fetcher. 
Using VCR gem

https://github.com/skywinder/github-changelog-generator/pull/381/commits/951a9c3a21ffa42ee3e9bdc7adfee642a91de095

awaage:
Added specs for the new octo_fetcher 
Commit ( https://github.com/skywinder/github-changelog-generator/pull/381/commits/434cc850ea05db847625845b8b6b92f310f460e3 )

I standardized the OctoFetcher to return hashes/arrays with stringified keys, as opposed to the Sawyer::Resource type. This was to keep it simple, and as similar as possible to old code / tests.

I had to refactor some of the code so that it uses hash access with string keys rather than accessing as methods/objects.

olleolleolle:
(@awaage, very excited about this! Super-cool work.)

awaage:
@olleolleolle thanks! I ran into some API limits testing on my big repository last night. Hope to get to some real testing this weekend!

awaage:
This PR is done

skywinder:
@awaage Nice! Please, run `rubocop -a` inside the project and commit corrections (to make syntax unified and make Travis checks pass https://travis-ci.org/skywinder/github-changelog-generator/builds/132438761 )

olleolleolle:
First thought: Can this change also remove the `github_api` gem as a dependency?

awaage:
@olleolleolle - the follow up is to remove Fetcher completely, which means remove the specs and the dependencies. I can do that now, but I left it initially for comparison purpose.

awaage:
I have created a new PR #384 to clean up, remove old fetcher, and make rubocop happy. Please see comments there for more information.

olleolleolle:
I like the changes, and I'm glad you decided to go to these lengths to make the most of this change.

This puts us in a better position to improve debug logging or use any other Faraday tools that are available.

olleolleolle:
@awaage Hello, again!

I got inspired to work further on this change, by extending your branch.

I'll make a Pull Request on top of `retentionscience:aw/octokit-fetcher`, so that you can review the changes in good time. 

Things I've done now: 

- drop github_api (you'd done a great job of preparing that)
- depend on multi_json directly (it was a dependency somewhere)
- upgrade colorize to 0.8.1 (to avoid deprecation warnings)
- use [VCR with its RSpec metadata](https://www.relishapp.com/vcr/vcr/v/3-0-1/docs/test-frameworks/usage-with-rspec-metadata) (add a `:vcr` just next to the `context`, before its `do` block declaration)

This had the effect that we got rid of:

- rack as a dependency (currently, Rack 2.x as a dependency stops Ruby < 2.2 from installing our gem)
- hashie as a dependency (currently working on stepping away from this in our specs, mostly deleting things)

Having fun!

olleolleolle:
Duplicate of #422.  Closing this issue. Thank you!

PR #383, opened 2016-05-22
Fetch commits on master.

inet-kamal:
Display commits in the change log done on master in their respective releases.


olleolleolle:
I think latest release is doing this? If not, please re-open this issue with a reproduction of the problem.

PR #385, opened 2016-05-29
Slack notification

inet-kamal:
Mandatory parameters:
--slack_notifier true --slack_webhook_url WEBHOOKURL
Optional:
slack_username & slack_channel, default values webhookbot and general

Ex: command:
 github_changelog_generator -u test -p test -t TOKEN --slack_notifier true --slack_webhook_url WEBHOOK_URL --slack_channel random --slack_username testUser

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source "https://rubygems.org"
 
 gemspec
 
+gem "slack-notifier"
 group :test do
   gem "rake"
   gem "bundler"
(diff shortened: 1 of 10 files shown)


olleolleolle:
Thanks so much for spending time on this amazing patch. We really appreciate it. However, I do not think this is something we want to add right now, since Slack notification can be done as a command-line invocation, on the content that we generate.

Perhaps we can add a README section on things one can do with the generated CHANGELOG.md when it's been produced, and then a "Send it to Slack" recipe, using another tool, for example [slackcat](https://github.com/vektorlab/slackcat).

Thanks again! I'll close this PR.

Issue #365, opened 2016-04-12
Pull requests seem to be inaccurately included in tags


dguo:
Hi, thanks for this tool. I have an issue regarding the pull request logic. My team follows the  [Git Flow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) workflow, and our changelog seems to be inaccurate as a result. The list of merged pull requests under each tag contains all the pull requests that were merged into any branch during the time between that tag and the previous tag, *regardless* of whether or not they are actually included in the tag. I essentially only want to see the pull requests that are associated with the commits that are displayed when I follow the `Full Changelog` link. Is that currently doable?

My team's project is not open source, so I can't provide its changelog as an example, but I'm happy to create a dummy project to reproduce the issue, if that would help.

skywinder:
Hi, there! Thanks for mention it.
To solve this issue it's required to find a way to check, is PR merged to `master` or not, Still Wondering how to do that.

dguo:
Hmm..what about [getting the list of commits in the PR](https://developer.github.com/v3/pulls/#list-commits-on-a-pull-request) and then checking if the tag contains those commits? Not sure how much that could affect performance.

skywinder:
@dguo thanks, it makes sense. I will check, how it works. It would be very nice improvement!

tamers:
I started using this awesome tool a week ago ( @skywinder thanks btw ) but I have the same exact problem as @dguo so +1 for this feature

dirk-proper:
+1 for this feature. We see duplicates in the changelog due to having more than one pull request from the same feature branch into multiple branches (develop and master).

dpieri:
Here's a similar problem:

All of our tags are on Master. Often we have long lived feature branches that will have multiple PRs against them (breaking up the feature into multiple PRs so that the discrete PRs are easier to review). When the feature is done it finally gets merged into master via a final PR.

With the current version of github_changelog_generator those incremental PRs (which are not yet in master) show up under tags as soon as they're merged into the feature branch.

Issue #304, opened 2015-11-12
since_tag seems to not be working


lamont-granquist:
I'm running with this config:

``` ruby
GitHubChangelogGenerator::RakeTask.new :changelog do |config|
  config.since_tag = 'v5.0.0'
  config.exclude_labels = %w[duplicate question invalid wontfix changelog_skip]
  config.issues = false
end
```

And expecting that what I'll get is the contents of my HISTORY.md file, and only the PRs since v5.0.0 (grouped under 'unreleased').  What I'm getting is all the PRs in the project with no tag orgnization to them at all.  If I throw in `since_tag = 'v4.0.0'` then I do get a correct section for the unreleased tags, but then the v5.0.0 tag gets all the rest of the PRs.

I tracked it down far enough that it looks like `#delete_by_tag` is being passed 'nil' for both the older_tag and newer_tag which seems pretty clearly wrong, but hit a timebox and didn't track it down any further.

lamont-granquist:
Link to PR with code + output of running 'rake changelog':

https://github.com/acrmp/foodcritic/pull/406

grEvenX:
Having the same issue

lamont-granquist:
So, and to be clear it looks like the `since_tag` works correctly for determining how many 'blocks' of PRs there are and their heading.  It also works for the top N-1 blocks/headings.  But the last block/heading gets all the rest of the PRs thrown in it and ignores the since_tag, effectively dumping all of the PRs into the changelog.

swarzech:
+1

lamont-granquist:
might be a dup of #282 ?

sciutand:
:+1:

skywinder:
@lamont-granquist  Hi. Is this issue still valid? Can you give me a name of your repo, if it's open sourced, in order to test it?

lamont-granquist:
still valid last i checked.

acrmp/foodcritic was the repo i was testing it against, the PR linked to in the 2nd comment shows the results, you should be able to check out that PR against foodcritic and replicate the output.

nrodriguez:
Still getting this issue with 1.14.3

alexjfisher:
I can't get since_tag to work either.  I've been trying to migrate https://github.com/voxpupuli/puppet-archive

alexjfisher:
[...]

Issue #282, opened 2015-09-21
PR appears as part of "unreleased


abonas:
Generating the changelog on latest master of this repo:
https://github.com/abonas/kubeclient/

Shows that the PR https://github.com/abonas/kubeclient/pull/118
is part of "unreleased" section, however it was released as part of v 0.5.1
as can be seen here:
https://github.com/abonas/kubeclient/compare/v0.5.0...v0.5.1

abonas:
Generating the changelog on latest master of this repo:
https://github.com/abonas/kubeclient/

Shows that the PR https://github.com/abonas/kubeclient/pull/118
is part of "unreleased" section, however it was released as part of v 0.5.1
as can be seen here:
https://github.com/abonas/kubeclient/compare/v0.5.0...v0.5.1

skywinder:
@abonas  Thanks. I will take a look next week, when come from vacation.

olleolleolle:
Hi, this is still a defect. I label this issue BUG now.

The tool puts PRs in the wrong bucket. 

In this small snippet, we can find PR 118 wrongly placed in the section generated after 0.5.1:

```
## [v0.6.0](https://github.com/abonas/kubeclient/tree/v0.6.0) (2015-09-21)
[Full Changelog](https://github.com/abonas/kubeclient/compare/v0.5.1...v0.6.0)

**Closed issues:**

- Parse Kubernetes config file to create clients [\#122](https://github.com/abonas/kubeclient/issues/122)
- Move kubeclient default version to kubernetes v1 [\#101](https://github.com/abonas/kubeclient/issues/101)

**Merged pull requests:**

- update default version of API to v1 instead of v1beta3 \[\#101\] [\#120](https://github.com/abonas/kubeclient/pull/120) ([jonmoter](https://github.com/jonmoter))
- check for missing item list returned due to k8b bug\[1\] [\#118](https://github.com/abonas/kubeclient/pull/118) ([moolitayer](https://github.com/moolitayer))
- move code that can be shared between clients to common [\#114](https://github.com/abonas/kubeclient/pull/114) ([moolitayer](https://github.com/moolitayer))
- Add k8s guestbook tests and get\_entities by namespace [\#105](https://github.com/abonas/kubeclient/pull/105) ([abonas](https://github.com/abonas))

## [v0.5.1](https://github.com/abonas/kubeclient/tree/v0.5.1) (2015-08-27)
[...]

PR #387, opened 2016-06-02
Fixes for Tag handling

awaage:
Main problem is that previously, it would put all PRs into the newest Tag section.
The reason is because the since_tag would be detected as nil, so it would include all tags from the beginning of time. I changed the logic to make a hash, with:
key: tag to make section
value: [left-tag, right-tag]

Now, we iterate through this hash, and create sections for each, choosing the PRs committed between left-tag date and right-tag date.

Related to:
https://github.com/skywinder/github-changelog-generator/issues/365

https://github.com/skywinder/github-changelog-generator/issues/304

https://github.com/skywinder/github-changelog-generator/issues/282

### DIFF
--- a/lib/github_changelog_generator/generator/generator_fetcher.rb
+++ b/lib/github_changelog_generator/generator/generator_fetcher.rb
@@ -12,13 +12,13 @@ module GitHubChangelogGenerator
     end
 
     # Async fetching of all tags dates
-    def fetch_tags_dates
+    def fetch_tags_dates(tags)
       print "Fetching tag dates...\r" if @options[:verbose]
       # Async fetching tags:
       threads = []
       i = 0
-      all = @filtered_tags.count
-      @filtered_tags.each do |tag|
+      all = tags.count
+      tags.each do |tag|
         print "                                 \r"
         threads << Thread.new do
           get_time_of_tag(tag)
--- a/lib/github_changelog_generator/generator/generator_generation.rb
+++ b/lib/github_changelog_generator/generator/generator_generation.rb
@@ -137,7 +137,7 @@ module GitHubChangelogGenerator
 
       log = generate_unreleased_section
 
-      @tag_section_mapping.each_pair do |tag_section, left_right_tags|
+      @tag_section_mapping.each_pair do |_tag_section, left_right_tags|
         older_tag, newer_tag = left_right_tags
         log += generate_log_between_tags(older_tag, newer_tag)
       end
--- a/lib/github_changelog_generator/generator/generator_generation.rb
+++ b/lib/github_changelog_generator/generator/generator_generation.rb
@@ -141,12 +141,6 @@ module GitHubChangelogGenerator
         older_tag, newer_tag = left_right_tags
         log += generate_log_between_tags(older_tag, newer_tag)
       end
-      # (1...filtered_tags.size).each do |index|
-      #   log += generate_log_between_tags(filtered_tags[index], filtered_tags[index - 1])
-      # end
-      # if @filtered_tags.count != 0
-      #   log += generate_log_between_tags(nil, filtered_tags.last)
-      # end
 
       log
     end
--- a/lib/github_changelog_generator/generator/generator_generation.rb
+++ b/lib/github_changelog_generator/generator/generator_generation.rb
@@ -5,7 +5,6 @@ module GitHubChangelogGenerator
     # @return [String] Generated change log file
     def compound_changelog
       fetch_and_filter_tags
-      sort_tags_by_date(@filtered_tags)
       fetch_issues_and_pr
 
       log = ""
@@ -138,12 +137,16 @@ module GitHubChangelogGenerator
 
       log = generate_unreleased_section
 
-      (1...filtered_tags.size).each do |index|
-        log += generate_log_between_tags(filtered_tags[index], filtered_tags[index - 1])
+      @tag_section_mapping.each_pair do |tag_section, left_right_tags|
+        older_tag, newer_tag = left_right_tags
+        log += generate_log_between_tags(older_tag, newer_tag)
       end
-      if @filtered_tags.count != 0
-        log += generate_log_between_tags(nil, filtered_tags.last)
-      end
+      # (1...filtered_tags.size).each do |index|
+      #   log += generate_log_between_tags(filtered_tags[index], filtered_tags[index - 1])
+      # end
+      # if @filtered_tags.count != 0
+      #   log += generate_log_between_tags(nil, filtered_tags.last)
+      # end
 
       log
     end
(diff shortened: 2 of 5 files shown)


lib/github_changelog_generator/generator/generator_fetcher.rb
```diff
@@ -12,13 +12,13 @@ def fetch_events_for_issues_and_pr
     end
 
     # Async fetching of all tags dates
```
awaage: Make this more functional, so we can track what's going on, instead of modifying the instance variable.

lib/github_changelog_generator/generator/generator_generation.rb
```diff
@@ -138,11 +137,9 @@ def generate_log_for_all_tags
 
       log = generate_unreleased_section
```
awaage: This is the main change. Using the hash with left-tag, right-tag to create sections

awaage: This is intentionally more explicit, than trying to iterate through the `@filtered_tags` and create sections from that array.

lib/github_changelog_generator/generator/generator_generation.rb
```diff
@@ -5,7 +5,6 @@ class Generator
     # @return [String] Generated change log file
     def compound_changelog
       fetch_and_filter_tags
-      sort_tags_by_date(@filtered_tags)
```
awaage: moved this method call to the `fetch_and_filter_tags` method

lib/github_changelog_generator/generator/generator_tags.rb
```diff
@@ -2,11 +2,33 @@ module GitHubChangelogGenerator
   class Generator
     # fetch, filter tags, fetch dates and sort them in time order
     def fetch_and_filter_tags
```
awaage: The logic should be pretty clear now. The steps are very serial.

1. detect the since_tag and due_tag up front, before any logic
2. then call fetcher to get all tags
3. then fetch tag dates
4. then sort by date
5. then filter tags
6. then build tag section mappings

lib/github_changelog_generator/generator/generator_tags.rb
```diff
@@ -69,6 +95,7 @@ def version_of_first_item
     # @return [Array]
     def get_filtered_tags(all_tags)
       filtered_tags = filter_since_tag(all_tags)
+      filtered_tags = filter_due_tag(filtered_tags)
```
awaage: Noticed `filter_due_tag` was not called previously. I think that is a bug, so I added it.

lib/github_changelog_generator/generator/generator_tags.rb
```diff
@@ -97,13 +124,13 @@ def filter_since_tag(all_tags)
     # @return [Array] filtered tags according :due_tag option
     def filter_due_tag(all_tags)
       filtered_tags = all_tags
-      tag = @options[:due_tag]
+      tag = detect_due_tag
       if tag
         if (all_tags.count > 0) && (all_tags.map(&:name).include? tag)
           idx = all_tags.index { |t| t.name == tag }
```
awaage: Simplified this logic a bit to make it easier to follow. No changes logically

skywinder:
@awaage Thanks! 
I see you just update PR 2 minutes ago. Is it ready?
Your logic looks good, but have to double check everything.

awaage:
@skywinder there's one issue with filtering excluded tags. That logic needs to happen inside`fetch_and_filter_tags` as well. I'll add that shortly.

awaage:
Ok, ready to go.

spec/unit/generator/generator_tags_spec.rb
```diff
@@ -64,7 +64,7 @@ def tags_mash_from_strings(tags_strings)
     context "with excluded and between tags" do
       let(:generator) { GitHubChangelogGenerator::Generator.new(between_tags: %w(1 2 3), exclude_tags: %w(2)) }
       it { is_expected.to be_a Array }
-      it { is_expected.to match_array(tags_mash_from_strings(%w(1 3))) }
+      it { is_expected.to match_array(tags_mash_from_strings(%w(1 2 3))) }
```
skywinder: Why you put tag `2` in expected result, when it's passed as excluded tag?

awaage: I changed the responsibility of get_filtered_tags.

Now, get_filtered_tags only does 3 things:
 - filtered_tags = filter_since_tag(all_tags)
 - filtered_tags = filter_due_tag(filtered_tags)
 - filter_between_tags(filtered_tags)


The " excluded tags " now happens in `fetch_and_filter_tags`.

olleolleolle: Changed responsibilities often mean changed tests, removed tests and most often changed test descriptions. 

If behavior changes, tests change with it. 

Before we drop or change a test, we want to know that we check that behavior where it happens (now). 

So, where do we want to read about the excluding tags bit? What behavior supports that behavior?

awaage: There's no tests for `fetch_and_filter_tags` which is the most important method.

IMO the current tests are kinda false sense of security, because in fact the code is not working as expected. 

I can add tests but will take a couple weeks. It will be MUCH easier to do this in my other branch, where I've set up VCR, etc. for testing. It doesn't really make sense to add the tests in this PR.

olleolleolle: After thinking a lot about of these changes, I like them.

The new thing is that **excluding tags is a top-level concern**. It's applied as a part of fetching the tags, stripping those excluded away at that point. The semantics on this lower level has changed. For the better – "confident code", as it were.

Here is a test "the old behavior" with this PR's code, which would take exclusions into account on each level:

```ruby
  describe "#get_filtered_tags" do
    subject do
      generator.filter_excluded_tags(
        generator.get_filtered_tags(tags_from_strings(%w(1 2 3 4 5)))
      )
    end

    context "with excluded and between tags" do
      let(:generator) { GitHubChangelogGenerator::Generator.new(between_tags: %w(1 2 3), exclude_tags: %w(2)) }
      it { is_expected.to be_a Array }
      it { is_expected.to match_array(tags_from_strings(%w(1 3))) }
    end
  end

```

The new semantics work like awaage noted:

```ruby
  describe "#get_filtered_tags" do
    subject do
      generator.get_filtered_tags(tags_from_strings(%w(1 2 3 4 5)))
    end

    context "respects between tags" do
      let(:generator) { GitHubChangelogGenerator::Generator.new(between_tags: %w(1 2 3)) }

      it { is_expected.to be_a Array }
      it { is_expected.to match_array(tags_from_strings(%w(1 2 3))) }
    end
  end
```

olleolleolle: I made the changes in the `context` text in https://github.com/retentionscience/github-changelog-generator/pull/2

olleolleolle:
Duplicate of #422.  Closing this issue. Thank you!

PR #391, opened 2016-07-02
Parser: YARD docstrings and a rename, and RegExp named capture groups

olleolleolle:
This PR improves **documentation** and very small refactoring: **extract two RegExps**, and give their capture groups labels.

While reading the code, I noted that there were a few additional methods that had quite little documentation. While documenting, I named a few variables.

Interested parties can use YARD to generate documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ pkg/
 coverage/
 .bundle
 spec/*.lock
+doc
+.yardoc
--- a/.yardopts
+++ b/.yardopts
@@ -1,0 +1,2 @@
+--markup=markdown
+
--- a/lib/github_changelog_generator/parser.rb
+++ b/lib/github_changelog_generator/parser.rb
@@ -23,7 +23,12 @@ module GitHubChangelogGenerator
       options
     end
 
-    # @param [Hash] options to display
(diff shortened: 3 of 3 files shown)


olleolleolle:
(That build failure is rack 2.x demanding 2.2+ of Ruby to work.)

skywinder:
@olleolleolle Perfect, as usual :)
Seems, there is no limitation about upper version of ruby. But have to check.
Can you bump ruby version in `.travis.yml` in order to make travis works?

olleolleolle:
If-and-when we get rid of the Rack dependency, we'd be able to ignore that bit, and be compatible farther down the EOL chain of Rubies.

BUT, we could also make a definitive call and say "the project supports stable-and-not-EOL'd Rubies".

skywinder:
@olleolleolle The only reason to don't do that - is support of old versions of ruby, am I right? 
Is so, I don't mind in support of outdated versions of ruby. So, let's do this. 
We can bump major number with this  move. So, anyone, who want to use this project - can stay on `1.x`. Versions.

olleolleolle:
@skywinder This PR does not belong in the 2.0.0 milestone. If you review the code changes in this PR, you can see nothing in this changed dependencies.

PR #399, opened 2016-07-18
Show how to use it with Rakefile

edusantana:
It was missing the require statement.


olleolleolle:
@edusantana  Thanks a lot!

PR #400, opened 2016-07-20
Don't constrain runtime deps.

jkeiser:
Since we are a development dependency of many projects, constraining our runtime deps can cause their Gemfiles to downgrade dependencies they share with us, which is never desirable.

Further, since these dependencies are < 1.0, there is no semver guarantee of backwards compatibility from 0.x to 0.y, so the constraint doesn't actually buy us anything.


olleolleolle:
Thanks, @jkeiser - this is a very thoughtful improvement of our package ecology situation. 

We want to be a good neighbour, and we want the best and most flexible Ruby package life for our users.

I'll merge this _now_.

skywinder:
Fully agree with @olleolleolle , thank you, @jkeiser!

hunner:
Rebased commit: 4969108eca0c15cffa441fb1b3b8490361b842b1

PR #401, opened 2016-07-21
Ability to implicity set user and project from command line

### LINKED ISSUE #397: `--user` flag should not be ignored
I am specifying a user using the `-u` or `--user` flag in the command line.

I expect this to result in my command line specified user being used.

The actual result is that it "detects" me as the user, and uses me instead.

The issue with this is that the repository I am trying to generate release notes for is not owned by me, I work from (and issue pull requests) from a fork of a repo owned by another "user" (which is where we manage releases from). So, when I'm trying to generate release notes, I need to hit that user instead.

```
github_changelog_generator --no-author --compare-link --github-site=https://github.mysite.com --token=$CHANGELOG_GE_TOKEN --github-api=https://github.mysite.com/api/v3/ -u=shareduser
Detected user:cormacmccarthy, project:your-project
...
:user=>"cormacmccarthy",
...
```

skywinder:
This fix #397

### DIFF
--- a/lib/github_changelog_generator/parser.rb
+++ b/lib/github_changelog_generator/parser.rb
@@ -61,7 +61,7 @@ module GitHubChangelogGenerator
         opts.on("-b", "--base [NAME]", "Optional base file to append generated changes to.") do |last|
           options[:base] = last
         end
-        opts.on("--bugs-label [LABEL]", "Setup custom label for bug-fixes section. Default is \"**Fixed bugs:**""") do |v|
+        opts.on("--bugs-label [LABEL]", "Setup custom label for bug-fixes section. Default is \"**Fixed bugs:**" "") do |v|
           options[:bug_prefix] = v
         end
         opts.on("--enhancement-label [LABEL]", "Setup custom label for enhancements section. Default is \"**Implemented enhancements:**\"") do |v|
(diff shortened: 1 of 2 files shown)


skywinder:
@olleolleolle please, kindly check

lib/github_changelog_generator/parser.rb
```diff
@@ -61,7 +61,7 @@ def self.setup_parser(options)
         opts.on("-b", "--base [NAME]", "Optional base file to append generated changes to.") do |last|
           options[:base] = last
         end
-        opts.on("--bugs-label [LABEL]", "Setup custom label for bug-fixes section. Default is \"**Fixed bugs:**""") do |v|
+        opts.on("--bugs-label [LABEL]", "Setup custom label for bug-fixes section. Default is \"**Fixed bugs:**" "") do |v|
```
olleolleolle: The string looks fishy.

Perhaps you can use `'Setup custom label for bug-fixes section. Default is "**Fixed bugs:** "'`?

skywinder: Oh, right. Long time ago we decide to go with double quotes only, to reduce mess. Anyway there is missed escape character. Thanks. Fixed! Anything else? 😉

olleolleolle:
Cool!

This code is _much_ easier to read. It returns values, as opposed to mutating passed-in values.

(To better read what was happening, I used the Github "ignore whitespace" `?w=0` on the diff URL.)

I am merging!

PR #405, opened 2016-08-09
Report actual github error when rate limit exceeded

jkeiser:
Also, don't report rate limit exceeded error unless the rate limit was exceeded.


jkeiser:
@olleolleolle note that this is different from the previous error message fix: it propagates the change to a bunch of other places.

olleolleolle:
Thanks! Merging!

PR #408, opened 2016-08-31
Replace GPL'd colorize gem with MIT-licensed rainbow

jamesc:
colorize is licensed under GPL-2 which conflicts with the MIT license of
github_changelog_generator.  This changes all usage of colorize to
rainbow which does have a compatible license (MIT)

### DIFF
--- a/github_changelog_generator.gemspec
+++ b/github_changelog_generator.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency "rake", ">= 10.0"
   spec.add_runtime_dependency "github_api", ">= 0.12"
-  spec.add_runtime_dependency "colorize", ">= 0.7"
+  spec.add_runtime_dependency "rainbow", ">= 2.1"
 
   spec.add_development_dependency "overcommit", ">= 0.31"
   spec.add_development_dependency "rspec", ">= 3.2"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     github_changelog_generator (1.13.1)
-      colorize (>= 0.7)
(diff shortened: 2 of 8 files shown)


olleolleolle:
Hello James! 

Good choice of library.

I wonder: we're coloring strings that pass through our `log` function in the `Helper` namespace. Idea: Can we ditch all spread-out coloring of these?

olleolleolle:
@jamesc Hi! I made my comment into a PR-onto-your-PR.

olleolleolle:
@jamesc Oh, I made a change, again.

github_changelog_generator.gemspec
```diff
@@ -27,6 +27,9 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "github_api", ">= 0.12"
   spec.add_runtime_dependency "rainbow", ">= 2.1"
 
+  # rack 2.0 does not work on Ruby < 2.2.  Lock it down here
+  spec.add_runtime_dependency "rack", "~> 1.2"
```
olleolleolle: This is contentious: if people are using this library in their application, we're holding the rack back.

jamesc:
Yeah, I backed out that change.  It doesn't fix the travis builds anyway.

olleolleolle:
@jamesc Cool, thanks for your hard effort!

This change fixes a hard problem, and we got the logging mechanism placed more centrally, by only using its formatter. And we got rid of the String extension, which the colorize gem also had.

To make the PR really stand out in the changelog, do you want to rephrase the title of it? Since we will be making a CHANGELOG.md out of this, something clearer (for understanding at a glance). Examples include:

- Use rainbow instead of colorize
- Replace GPL'd colorize gem with MIT-licensed rainbow

olleolleolle:
Again, thanks a lot for doing this, @jamesc! 🎉

PR #417, opened 2016-09-22
Travis: Collecting the config, gemspec: extract development deps to Gemfile

olleolleolle:
This PR wants to make it easier to run builds, and understand what's running where.

Travis configuration:

- Comment on what the Travis configuration does
- Travis configuration: tighter matrix, each of the Rubies used has its own function:
  - 2.2.2 is the earliest Ruby we support out of the box: the build-and-install tests are its bailiwick
  - 2.1 has a Gemfile which holds back rack to 1.6 (the README includes a Troubleshooting FAQ item about "My Ruby is old, can I still use this?" which explains Gemfiles)
  - 2.3.1 has a Gemfile which does not hold rack back

The gemspec got these changes:

- Place development dependencies in the Gemfile
- Upgrade the Github API gem to 0.14 - this avoids warnings


olleolleolle:
Hello, @jkeiser! I would like to hear your thoughts on this set-up.

Does it go far enough?

PR #420, opened 2016-09-27
Travis: Test on 2.4.0-preview2

olleolleolle:
This PR adds another build to the rvm matrix: 2.4.0-preview2.

It also removes and git ignores the Gemfile.lock, so that each build of a gem comes from a basic gemfile.


skywinder:
You're amazing, @olleolleolle . Thank you for support!

PR #384, opened 2016-05-25
Original Fetcher removed and rubocop passes

awaage:
Two things:
1. Original Fetcher removed 
2. rubocop passes

This PR was branched off of (https://github.com/skywinder/github-changelog-generator/pull/381) 

So it doesn't clutter the changeset of (https://github.com/skywinder/github-changelog-generator/pull/381)

This current PR is the FULL changeset, inclusive of #381.

My advice is to review changeset of #381 first, and then instead of merging that one, just merge This one (#384) which will handle test passing / rubocop passing / removing original Fetcher.

Hope that makes sense.

PR #382, opened 2016-05-18
Octokit Caching Github API requests - WIP

awaage:
This is still a WIP.

This changeset should actually be much smaller.
This was branched off of a previous PR (https://github.com/skywinder/github-changelog-generator/pull/381/files)

Once that one is approved, this changeset will appear much smaller.

Need to:
- Figure out how to parameterize the settings for cache
- Not necessarily require activesupport gem..

But, good news is **it works!**

Addresses: https://github.com/skywinder/github-changelog-generator/issues/361

Welcome feedback.

PR #422, opened 2016-09-27
Use octokit, carrying awaage commits

olleolleolle:
This is a rebased PR of this collection of changesets by @awaage:

- #381 Use OctoKit
- #382 Cache OctoKit GH requests
- #384 Replace Fetcher with OctoFetcher  
- #387 Tags handled differently

The goal is to have an **OctoKit**-based metadata fetching.

By using the Faraday-based OctoKit, we now have [VCR](https://relishapp.com/vcr/vcr/docs) support in the test suite, as well.

This PR represents the work of the Project card https://github.com/skywinder/github-changelog-generator/projects/1#card-278430


awaage:
Thanks @olleolleolle ! This is great!

awaage:
From this point, what else do we need? How do you plan to release this?

olleolleolle:
Did the public API change? If yes, I think we must bump the major version.

Do we stand to gain anything by keeping up a 1.x release series around? (Perhaps the freedom to innovate in the 2.x and cut off old platforms?) A way to have the possibility to release patch fixes to 1.x versions. 

@skywinder Any thoughts/reactions? I'm curious to hear your ideas and hopes. Did you see the Project in the Github site? The one where I ask us to find an interface that a backend like OctoFetcher can implement. Would it enable a Gitlab user to build their own Fetcher, supporting other ways of providing data for the generator? 

Thanks for participating in all this.

olleolleolle:
Let's move this forward!

skywinder:
@olleolleolle That's pretty interesting idea with OctoFetcher! It would be amazing to extend all of this algorithms for Bitbucket too. I will think about that.

And in addition to my personal email:

As for me is very important to see, how the project is rising. 
And it could be real only your effort! 
Thank you very much for the help! 👍

olleolleolle:
Thank you for the trust and encouragement.

PR #427, opened 2016-09-30
Rubocop TODO file regenerated

olleolleolle:
This PR used Rubocop's auto-regenerate feature to keep the TODO file for linting updated.

The goal is to get rid of it altogether. We are making good progress!


skywinder:
Wow! Amazing!

PR #432, opened 2016-10-05
OctoFetcher, Options: Refactoring

olleolleolle:
This PR is a collaboration by @til, @bestie and @olleolleolle.

- extract Array, Hash stringifying extension methods
- add a class Options which protects against unknown/misspelled option usage
- dropped a spec marked pending, but which was really a TODO


lib/github_changelog_generator/options.rb
```diff
@@ -0,0 +1,91 @@
...
+      :unreleased_only,
+      :user,
+      :usernames_as_github_logins,
+      :verbose
+    ]
+
+    THESE_ARE_DIFFERENT = [
+      :tag1,
+      :tag2
+    ]
```
til: But _what_ are they? :smile:

olleolleolle:
I made an invocation of the tool on this repo, which has ~193 events to fetch. It's taking its sweet time doing that. I am very curious if this laggard speed is about our changes.

PR #434, opened 2016-10-08
On OctoKit::Forbidden error: retry with exponential backoff

awaage:
# Description:
For large repos, we sometimes hit the Github Forbidden exception. This currently does not retry, so the return value is nil. This causes other issues down the road, when dealing with the result of the API call. This fix will add sleep + retries to the call. The sleep time uses exponential back-off to decide on how long to sleep.


olleolleolle:
Thank you for the PR, I'll get this looked at very soon.

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -280,19 +281,47 @@ def iterate_pages(client, method, *args)
...
+    def sys_abort(msg)
+      abort(msg)
+    end
+
+    # Returns the exponential backoff (seconds) for this attempt number
+    #
+    # @param [Integer] attempt the attempt number
+    # @return [Integer] Exponential backoff seconds
+    def exp_backoff(attempt)
+      (2**attempt - 1) / 2
```
skywinder: Why did you decide to take an exponential algorithm and `MAX_FORBIDDEN_RETRIES = 100`? Is there some rule or it's just approximation?
 I just read about 401 error on https://developer.github.com/v3/#failed-login-limit and checked https://developer.github.com/guides/best-practices-for-integrators/
But didn't found any suggestions about the delay before next attempt.

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -280,19 +281,47 @@ def iterate_pages(client, method, *args)
...
       end
       value
     end
 
+    def sys_sleep(seconds)
+      sleep(seconds)
+    end
+
+    def sys_abort(msg)
+      abort(msg)
```
skywinder: What the reason to wrap single lines methods to another method? Is it required for tests below?

awaage: This was done for specs. I don't love it, but didn't know a better way

lib/github_changelog_generator/octo_fetcher.rb
awaage: Good question. I added this later and didn't think too much about the MAX. Just thought it's better to not do this indefinitely.

awaage: You're right. It's not very explicit in their API docs. It's actually under the API limit, but Github flags certain requests that match their "abuse detection". It happens frequently for me (large git repo) when fetching tag-dates. I believe this is a simple API call, that happens very rapidly, 100s of times in a row. Github sees this as potentially abuse, so they give the `Octokit::Forbidden` error.

olleolleolle:
I have thought a bit, and looked around the Web for extracted retry mechanisms that have the exp backoff.

There are gems that do this, all with different feature-sets.

We need: exponential backoff (perhaps randomized) and the ability to react differently to differents kinds of exception.

In a future where we have more than 1 kind of fetcher, it would be nice to separate out the algorithm for retrying-and-backing-off from the API interactions themselves.

Can we implement the same functionality using (for example) https://github.com/ooyala/retries ?

awaage:
Seems like a good idea. 

That gem you posted looks ok and is pretty simple.

I have used this one before personally: https://github.com/kamui/retriable

olleolleolle:
@awaage Cool gem, `retriable`, that's a better suggestion. It also has no other dependencies.

awaage:
@olleolleolle What is the process for adding new gem? I need to add to all Gemfile versions?

olleolleolle:
Use the gemspec for direct dependencies like this. 

Gemfiles inherit those.

olleolleolle:
This is getting good.

- Rubocop 0.44.1 is complaining about style, but most of them you should be able to auto-get-rid-of: `rubocop -a lib/github_changelog_generator/octo_fetcher.rb`
- One new Rubocop complaint is "block has too many lines", and there I suggest adding `# rubocop:disable Metrics/BlockLength` as an end-of-line comment on the line that it complains about (this was in OptionParser).

awaage:
@olleolleolle updated for rubo ! thx

olleolleolle:
Thank you, @awaage for taking this to completion.

PR #438, opened 2016-11-04
Fixed issue #304 - entries of previous tags are included.

thorsteneckel:
This pull requests fixes the issue #304 .

ATTENTION: This pull request changes the current behavior!

Current behavior:
Use since_tag or due_tag and every issue, PR etc. that happened before the last tag will get added to the last tag.

New behavior:
Use since_tag or due_tag and every issue, PR etc. that happened in the timeframe of the tags is added to the tags. Previous elements are not added to the list.

Feedback is welcome. Cheers!

### DIFF
--- a/lib/github_changelog_generator/generator/generator.rb
+++ b/lib/github_changelog_generator/generator/generator.rb
@@ -11,7 +11,7 @@ module GitHubChangelogGenerator
   end
 
   class Generator
-    attr_accessor :options, :filtered_tags, :github, :tag_section_mapping
+    attr_accessor :options, :filtered_tags, :github, :tag_section_mapping, :sorted_tags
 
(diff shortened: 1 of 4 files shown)


thorsteneckel:
@olleolleolle I thought about adding tests and stuff but since there were none checking the current behavior I skipped them too. Imho there are some bigger changes on the structure necessary to test this properly. My suggestion would be to add a layer that makes it possible to mock github calls completely.

olleolleolle:
Does this change in behavior mean that options' descriptions need to change? Any other documentation around `due_tag`, `since_tag`?

olleolleolle:
Update: I've now re-read the relevant issue that this fixes. It's great!

thorsteneckel:
Wow! thanks @olleolleolle for merging this that fast.

PR #451, opened 2016-11-13
Drop Project-and-Username-finding code

olleolleolle:
~This PR is a refactoring to extract a large section of unrelated-to-option-parsing code from the option parser.~

~- [ ] improve extracted class name - is RepoSlugParser better?~
~- [ ] pore over the specs, esp. the final one, which seems fishy~
~- [ ] perhaps find a better way to express the "list of strategies" in the new class (`#call`)~
~- [ ] turn the default user/project repo slug into some kind of test setup~

**Update:** We did none of those things, we looked long and hard at it, and decided that it would be easier to drop the functionality.


lib/github_changelog_generator/project_name_finder.rb
```diff
@@ -0,0 +1,120 @@
+# frozen_string_literal: true
+module GitHubChangelogGenerator
+  class ProjectNameFinder
```
til: I find that trying to write a class level comment often helps with finding a fitting name. What purpose does this class serve, how does one use it?

lib/github_changelog_generator/project_name_finder.rb
```diff
@@ -0,0 +1,120 @@
...
+          puts "Can't detect user and name from first parameter: '#{args.first}' -> exit'"
+          return
+        end
+        if param
+          return NO_MATCHING_USER_AND_PROJECT
+        else
+          [Regexp.last_match(:user), Regexp.last_match(:project)]
+        end
+      end
+    end
```
til: Forgive me if I've overlooked something, since I'm only looking at this diff right now - but is this method and the following ones only public in order to support being called from the spec? In that case, it might be worth some effort trying to make them private I think.

olleolleolle: I think so, too. This is not a part of the usable interface.

lib/github_changelog_generator/project_name_finder.rb
olleolleolle: > This class interrogates given CLI options and Git remotes to parse out a GitHub username and a project name, the two parts of what's usually called a GitHub _repo slug_.

til: I see, makes sense. Sounds a bit like it does two different things that are not even so related: interrogating options on the one hand, parsing a remote on the other hand. Only the thing that comes out of it is the same.

olleolleolle: I'm also not excited about placing the test data (the default) inside this implementation.

I'll add "turning the default user/project repo slug into some kind of test setup".

olleolleolle: And I agree with this being two things.

olleolleolle:
Coming up for air after a long dive into this, I now find that I only want to remove this section of the code-base and have an easier time.

It's even easier to explain the thing when the `:github_site` option's been removed. 

I will update this PR with changes to that effect.

olleolleolle:
Ah, like a fine wine, this change has matured over the long summer.

Now I have made the change that I learned I wanted: code deletion.

hunner:
@olleolleolle Is this considered a breaking change from the 1.14.x series? Should the next release be a 2.0.0 instead of a 1.15.0 (not counting the alpha/beta/pre releases)? (There are many dramatic changes, but this may be the only breaking change that isn't technically a bugfix.)

olleolleolle:
If we could reimplement the feature with new code, perhaps we stand a better chance of making it work well?

Issue #450, opened 2016-11-10
What is the expected behavior of "between-tags"?


tduffield:
The wording of the description of "between-tags" says that the "Change log will be filled only between specified tags". What I took that to mean was that if I specified tags, the section headers would combine the changelog for those tags. 

For example, say I have the following tags: 1.2.[0-20], 1.3.[0-40], and 1.4.[0-10]. 
Now, lets say that I specify the following config: 
```
since-tag=1.2.0
between-tags=1.2.12,1.2.34,1.3.15,1.4.10
```

I would expect the following:

```
# 1.4.10
compare/1.3.16...1.4.10

# 1.3.15
compare/1.2.35...1.3.15

# 1.2.34
compare/1.2.13...1.2.34

# 1.2.12
compare/1.2.0...1.2.12
```

But instead, I get the following:

```
# 1.4.10
compare/1.4.9...1.4.10

# 1.3.15
compare/1.3.14...1.3.15

# 1.2.34
compare/1.2.33...1.2.34

# 1.2.12
compare/1.2.11...1.2.12
```

Which of these two is the expected behavior?

tduffield:
Sorry @olleolleolle, but it'd be great if I can have an answer in the next couple days. I have some time-sensitive work that depends on the answer.

olleolleolle:
I have not investigated `--between-tags`. Not sure what it wants to do. The actual behavior seems not-so-useful, though.

@skywinder, do you know? 

@tduffield If you feel like poking through the history and deduce, it's all there.

tduffield:
@olleolleolle I did look through the history briefly but I didn't see anything that answered my question super clearly.

skywinder:
The `--between-tags` not pretty useful and quite outdated option.
It works like `filter` for tags. i.e. leave tags, that specified after this option.
I think we can remove it in next major release to simplify the code.

tduffield:
I have to disagree. The between-tags is super useful for us (given my PR) in that we bump the version after every PR automatically but want the change log to reflect stable "releases." Im down if we want to deprecate that but having a way to specify section groupings is a must for us.

Lucashuang0802:
The `--between-tags` will be removed on the next release, @olleolleolle so this is not an issue anymore? 😃

PR #453, opened 2016-11-20
Fix section mapping, hiding untagged PRs, and hiding untagged issues

tduffield:
So this PR turned into a bit of a beast, but the foundation of it is to get three pieces of functionality working as they are documented. 

The first is the section mapping. I've included some tests to reflect the desired behavior but there are a few key changes. 

1. When `between-tags` is specified, it no longer leaves an empty section header. This was causing duplicate headers to show up when using base. Instead, the first/earliest between-tags tag will be included in the appropriate compare link. 
2. `between-tags` will now aggregate changes between the specified tags (as noted in #450). 
3. For the section header of the very last tag (the first tag to be created in the repo), the compare link will now (instead of not existing) show a link to compare the last tag against the initial commit. You can see this behavior in the tests reflected by a `nil` for the older_tag value. 

Next, I don't think that `add_issues_wo_labels` and `add_pr_wo_labels` were actually implemented. When I ran the utility against changelog_test, it would still show unlabeled PRs and Issues in the changelog even when those two values were set to false. This has been addressed.


thorsteneckel:
HI @tduffield - I like your commit. Actually we had the same issue in our company last week and I was fixing and debugging around trying to get the `add_issues_wo_labels` option working as expected. I found out that the `add_issues_wo_labels` option is only used when `include_labels` and/or `exclude_labels` are used, too.
Since I'm not deep enough into this I kept everything as it was and used a combination of these options. You approach is imho the right one. But since I'm not sure about it I'm really looking forward to a response of someone with deeper knowledge.

lib/github_changelog_generator/generator/generator.rb
```diff
@@ -55,22 +55,31 @@ def encapsulate_string(string)
     # @param [Array] pull_requests List or PR's in new section
     # @param [Array] issues List of issues in new section
     # @param [String] newer_tag Name of the newer tag. Could be nil for `Unreleased` section
-    # @param [String] older_tag_name Older tag, used for the links. Could be nil for last tag.
+    # @param [String] older_tag Older tag, used for the links. Could be nil for last tag.
```
olleolleolle: Doc block note: the older_tag is no longer a String, it's `[Hash,nil]`.

olleolleolle:
Your added tests will support lots of refactoring, which was less possible before you came along. 

This is great. 

Thank you!

wilson208:
I was experiencing issues with un-labeled PR's being included in the changelog and can confirm I have tested this PR and it addresses the issues I was seeing. Thanks a lot @tduffield!

Any idea of when/if this will be merged and released?

tphoney:
Is there anything i can do, to help get this merged ?

HAIL9000:
Hey again! Just wanted to check and see if there was anything I could do to help this PR along, we're currently blocked by this fix.

Thanks!

hunner:
This can be closed; merged in #550

olleolleolle:
Thanks for contributing this!

Closing this issue means progress was made, and you started that.

PR #455, opened 2016-11-22
Use Octokit::Client for both .com and Enterprise

eliperkins:
[`Octokit:EnterpriseAdminClient`](https://github.com/octokit/octokit.rb/blob/9fc209155dd4fcbf9595eb864df2276d36170a65/lib/octokit/enterprise_admin_client.rb#L12-L19) is only meant to be used for GHE admin
panel endpoints, such as Admin Stats, Management Console, etc.

This repairs GHE functionality by passing the API endpoint parameters
into `Octokit::Client` and letting it handle the rest.


olleolleolle:
👌Every day I'm learning. 

Can you run Rubocop with -a? That's the current Travis failure, easy to fix.

eliperkins:
🍏

tvon:
Would you mind cutting a release with this in it?  It seems to be a required change to work with GHE.

PR #470, opened 2017-01-07
Updated license year

kartikeys98:


skywinder:
According this http://webmasters.stackexchange.com/a/8249/44569
 I don't think that we need this changes.

Issue #475, opened 2017-02-08
SSL Cert issue on Windows


Arcanemagus:
With the fixes from #459 in place so that the program will run at all on Windows, we then start running into the following issue:

```raw
C:/Ruby23-x64/lib/ruby/2.3.0/net/http.rb:933:in `connect_nonblock': SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed (Faraday::SSLError)
        from C:/Ruby23-x64/lib/ruby/2.3.0/net/http.rb:933:in `connect'
        from C:/Ruby23-x64/lib/ruby/2.3.0/net/http.rb:863:in `do_start'
        from C:/Ruby23-x64/lib/ruby/2.3.0/net/http.rb:852:in `start'
        from C:/Ruby23-x64/lib/ruby/2.3.0/net/http.rb:1398:in `request'
        from C:/Ruby23-x64/lib/ruby/2.3.0/net/http.rb:1156:in `get'
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/faraday-0.10.0/lib/faraday/adapter/net_http.rb:81:in `perform_request'
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/faraday-0.10.0/lib/faraday/adapter/net_http.rb:41:in `block in call'
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/faraday-0.10.0/lib/faraday/adapter/net_http.rb:88:in `with_net_http_connection'
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/faraday-0.10.0/lib/faraday/adapter/net_http.rb:33:in `call'
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/faraday-0.10.0/lib/faraday/response.rb:8:in `call'
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/faraday-http-cache-2.0.0/lib/faraday/http_cache.rb:290:in `fetch'
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/faraday-http-cache-2.0.0/lib/faraday/http_cache.rb:195:in `process'
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/faraday-http-cache-2.0.0/lib/faraday/http_cache.rb:142:in `call!'
[...]

PR #480, opened 2017-02-08
bundled cacert.pem with --ssl-ca-file PATH option

olleolleolle:
This PR bundles a curl.haxx.se-provided `cacert.pem`, and configures Octokit to use it to verify its connections. If the user has configured SSL_CA_FILE or pointed at a pem path from the command-line, that default pem is not used.

This would help Windows users use the tool without surprises.

- [x] Name the option and document it
- [x] Rake task `:update_ssl_ca_file` which downloads from curl.haxx.se
- [x] Add a way for it to first read ENV variable `ENV["SSL_CA_FILE"]` if they exist
- [ ] ~Add the `:update_ssl_ca_file` to release routine~ (Post-poned: We need a written routine first.)
- [ ] ~Find a Windows + Ruby developer to prove this thing on their machine.~ (People have tried.)

See #475.


Arcanemagus:
I can possibly test this, but in no way would I call myself a Ruby developer 😛.

CDAGaming:
Tested this today in a sync on my fork, worked and generated perfectly no issues :)

PR #499, opened 2017-05-14
Fix RubyCops warnings for Travis CI

Lucashuang0802:
mostly there are three changes:
- change from `%w()` -> `%w[]`
- change from '%i()' -> `%i[]`

Also, there are twp weird warnings from `spec/`:
```
➜  github-changelog-generator git:(travis-ci-fix) ✗ bundle exec rake
Running RuboCop...
Inspecting 27 files
........................W..

Offenses:

spec/unit/parse_file_spec.rb:17:9: W: Lint/AmbiguousBlockAssociation: Parenthesize the param change { options } to make sure that the block will be associated with the change method call. (https://github.com/bbatsov/ruby-style-guide#syntax)
        expect { parser.parse! }.to_not change { options }
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/unit/parse_file_spec.rb:56:1: C: Style/IndentHeredoc: Use 2 spaces for indentation in a heredoc by using <<~ instead of <<. (https://github.com/bbatsov/ruby-style-guide#squiggly-heredocs)
exclude-labels=73a91042-da6f-11e5-9335-1040f38d7f90,7adf83b4-da6f-11e5-ae18-1040f38d7f90 ...
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

27 files inspected, 2 offenses detected
RuboCop failed!
```
I exclude it but this is opened for discussion. I guess running functional test cases are more important than style check.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,7 @@ AllCops:
   Exclude:
     - 'vendor/**/*'
     - 'gemfiles/**/*'
+    - 'spec/**/*'
 
 Metrics/LineLength:
   Enabled: false
--- a/Rakefile
+++ b/Rakefile
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 require "bundler"
 require "bundler/gem_tasks"
 require "rubocop/rake_task"
@@ -11,7 +12,7 @@ RuboCop::RakeTask.new
 RSpec::Core::RakeTask.new(:rspec)
 
 task :copy_man_page_to_manpath do |_t|
-  known_manpath_paths = %w(/etc/manpath.config /etc/manpaths)
+  known_manpath_paths = %w[/etc/manpath.config /etc/manpaths]
   manpath = known_manpath_paths.find do |f|
     path = Pathname(f)
     path.file? && path.readable?
@@ -36,5 +37,5 @@ task :copy_man_page_to_manpath do |_t|
   end
 end
 
-task checks: [:rubocop, :rspec]
-task default: [:rubocop, :rspec]
+task checks: %i[rubocop rspec]
+task default: %i[rubocop rspec]
--- a/lib/github_changelog_generator/generator/generator_processor.rb
+++ b/lib/github_changelog_generator/generator/generator_processor.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 module GitHubChangelogGenerator
   class Generator
     # delete all labels with labels from options[:exclude_labels] array
@@ -132,8 +133,8 @@ module GitHubChangelogGenerator
     # @return [Array] issues without labels or empty array if add_issues_wo_labels is false
     def filter_wo_labels(issues)
       if options[:add_issues_wo_labels]
-        issues_wo_labels = issues.select do |issue|
-          !issue["labels"].map { |l| l["name"] }.any?
+        issues_wo_labels = issues.find_all do |issue|
+          !issue["labels"].map { |l| l["name"] }.one?
         end
         return issues_wo_labels
       end
@@ -197,7 +198,7 @@ module GitHubChangelogGenerator
         end
       end
 
-      pull_requests.select! do |pr|
+      pull_requests.find_all! do |pr|
         !pr["merged_at"].nil?
       end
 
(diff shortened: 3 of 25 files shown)


Lucashuang0802:
Chnages on `lib/github_changelog_generator/generator/generator_processor.rb` needs more attention. 😃

.rubocop.yml
```diff
@@ -6,6 +6,7 @@ AllCops:
   Exclude:
     - 'vendor/**/*'
     - 'gemfiles/**/*'
+    - 'spec/**/*'
```
Lucashuang0802: This is also an important change but I am opened for discussion. 😃

olleolleolle: I think it's reasonable. Let's be careful about formatting and such, but not wasteful.

Rakefile
```diff
@@ -36,5 +37,5 @@ task :copy_man_page_to_manpath do |_t|
   end
 end
 
-task checks: [:rubocop, :rspec]
-task default: [:rubocop, :rspec]
+task checks: %i[rubocop rspec]
+task default: %i[rubocop rspec]
```
olleolleolle: This shines a light on an issue – but about this code: the symbol list notation was not introduced in 1.9.3. Our .gemspec notes 1.9.3 as the required Ruby version.

We do not _test_ on 1.9.3. It is EOL'd.

We still keep `- rvm: 2.1` in the Travis CI matrix. That is also EOL'd.

So, two things I think we should do, in other PRs:

- Drop `2.1` from the CI matrix
- Bump up required version in gemspec to something newer, like `">= 2.0.0"`, so that we can later merge _this_ PR. 
- Add a `TargetRubyVersion` to the .rubocop.yml, which points at the lowest tested version (currently: `"2.2.7"`)

lib/github_changelog_generator/generator/generator_processor.rb
```diff
@@ -132,8 +133,8 @@ def include_issues_by_labels(issues)
     # @return [Array] issues without labels or empty array if add_issues_wo_labels is false
     def filter_wo_labels(issues)
       if options[:add_issues_wo_labels]
-        issues_wo_labels = issues.select do |issue|
-          !issue["labels"].map { |l| l["name"] }.any?
+        issues_wo_labels = issues.find_all do |issue|
+          !issue["labels"].map { |l| l["name"] }.one?
```
olleolleolle: This seems like a turning-around of the PreferredMethods list:

https://github.com/bbatsov/rubocop/blob/master/config/default.yml#L670-L683

lib/github_changelog_generator/generator/generator_processor.rb
```diff
@@ -197,7 +198,7 @@ def filter_merged_pull_requests(pull_requests)
         end
       end
 
-      pull_requests.select! do |pr|
+      pull_requests.find_all! do |pr|
```
olleolleolle: Noting this, too. `select!` ought to be the right name for this.

olleolleolle:
@Lucashuang0802 Did you see the notes I made about possible PRs you could do?

Lucashuang0802:
@olleolleolle looks like I forgot to press the Comment button. Yeah, let's divide it into a couple PRs. 😃

Rakefile
Lucashuang0802: Oddly, it failed on both `2.3` and `2.4`. Would this change be contributing to fixing these two versions? @olleolleolle

.rubocop.yml
Lucashuang0802: 👍

olleolleolle:
Thank you for your PR!

I'll close this as master had moved on.

Lucashuang0802:
@olleolleolle thanks for this check-in: https://github.com/skywinder/github-changelog-generator/pull/516. Let me see what else to do to fix the CI. 💯

olleolleolle:
@Lucashuang0802 We're green; you should be able to view the board at: https://travis-ci.org/skywinder/github-changelog-generator/branches

Lucashuang0802:
👍 oh, that's nice! @olleolleolle

PR #514, opened 2017-05-22
Sync for fix/ssl-default-pem

CDAGaming:
Synced Changes from Master


olleolleolle:
I take it these changes were of an experimental character.

I merged the pem PR to master, so, I'll close this PR.

CDAGaming:
👍 Well, these changes were removing specific versions from those files so it would always target latest @olleolleolle , but i can remake this pr to be compatible with master branch soon

olleolleolle:
The version specifiers used are there to ensure we do not accept any gems that are _too old_.

Are any of them bothering your builds?

CDAGaming:
@olleolleolle No, they aren't bothering my builds, though some of the gems could be updated as some are not on latest versions(Some a wee bit old), so i could adjust my New PR to target the more recently released Gems

PR #515, opened 2017-05-22
Dependencies Update

CDAGaming:
Remove Specific Versioning Targets to always target latest versions & Update to Latest Ruby Version


CDAGaming:
Closing to keep in sync with master aka swapping branches

PR #519, opened 2017-05-25
Updates Dependencies (REUPLOAD!)

CDAGaming:


CDAGaming:
Ok, so This PR is Ready to Merge BUT there may be an incompatibility between 'codeclimate-test-reporter' and 'simplecov' but i can't be entirely sure. IF it does end up being an issue in the future, ill make a new PR after this to find compatible working versions

olleolleolle:
Hello!

Thank you for working to improve this project.

I am going to close this issue, and I'm going to explain why.

- The changes to the **Gemfile** affect what runs in **Travis CI** and on **contributors' computers**. Those are mostly harmless, and quite visible when they fail. Easy to spot and fix.
- The changes to the **gemspec** affect **users' projects**. If those projects put `github-changelog-generator` in their Gemfile (for example, to use Rake task method of configuration), then these pinned runtime dependencies will force versions upon them. _They_ may not need ActiveSupport 5. No need forcing that.

Reading material about Rubygems version specifiers: this is a long struggle.

- https://www.devalot.com/articles/2012/04/gem-versions.html
- http://yehudakatz.com/2010/08/21/using-considered-harmful-or-whats-wrong-with/

Thanks again for your hard work.

PR #526, opened 2017-06-03
Remove all old credit lines in the output then add a new one

### LINKED ISSUE #507: Bug: Credit line about this project added more than once
Generated changelog files get  a "credit line" added at the end.

Such a "credit line" about this project **should only be added once**. Currently, it is always appended, whether needed or not.

Here is what one of those lines look like:

```
\* *This Change Log was automatically generated by [github_changelog_generator](https://github.com/skywinder/Github-Changelog-Generator)*
```

### Desired behavior

- Avoid adding more lines like it if it already exists
- If such a string (or strings, there can have been many) can be found at the end of the generated file, delete it before adding a new one

### Code to work with

Where to find the code that does it: https://github.com/skywinder/github-changelog-generator/blob/d551c3a1987f34f6d1b20e2c35bcdf52b9a827ab/lib/github_changelog_generator/generator/generator_generation.rb#L23

### Example in the wild

https://github.com/lynndylanhurley/devise_token_auth/commit/5be8770e1e43eeeedc2d94e89d24f6e8d902d0b5

Enether:
Removes all old credit lines in the log and adds a new one.

Fixes #507 (Edit: added this annotation)


Lucashuang0802:
@Enether @olleolleolle i guess #535 is more readable? I am opened to discuss. 😄

Enether:
@Lucashuang0802 
Desired behavior was:

- Avoid adding more lines like it if it already exists
- If such a string (or strings, there can have been many) can be found at the end of the generated file, delete it before adding a new one
As far as I can tell, yours doesn't cover the second point. Also, `if log.include? credit_line == false` is unclear in my opinion, there should be parentheses for maximum readability.

@olleolleolle What gives, why hasn't my pull request been reviewed for 25 days while his got reviewed in no time?

Lucashuang0802:
@Enether I think it is probably too many things that make the PR got lost. I just noticed the PR this morning. But, I would love to use your solution if that makes the case. 👍

Lucashuang0802:
@Enether As for the second point, why don't you think it does not cover? The logic from my PR is: if it does not contain, do nothing. So you do not have to remove and add.

Lucashuang0802:
@Enether You are indeed correct. 👍  And we fixed it in #542.

Enether:
@Lucashuang0802 What are you talking about? You linked #542, which is something completely unrelated regarding a typo. What the original issue describe were two points, the second of which 

> If such a string (or strings, there can have been many) can be found at the end of the generated file, delete it before adding a new one

This means, that if a couple of strings are at the end of the file (through previous mis-generations, before this issue was fixed), they would stay there. What the second point wants is to remove them and have only one stay at the end.

Lucashuang0802:
You made a good point on the previous mis-generations thing. But, would the program generate a new CHANGELOG file each execution? If so, I guess we don't need `gsub!`. Otherwise, yes your solution would solve the previous mis-generations thing.

Lucashuang0802:
@Enether I have tried to manually duplicate credit lines and run the program again. It looks like it overrides the current one. So, we don't need to concern previous mis-generations.

olleolleolle:
@enether Thanks for the contribution. 

It looks grand and being the exact thing that I wrote about when I had the brainwave to put it in an Issue. It's an improvement. 

I want it in. Currently my GitHub shows this PR as having conflicts. Can you solve them? If yes, please do that. If no, let me know and I'll do it.

PR #535, opened 2017-06-27
Credit line bug

Lucashuang0802:
Fix issue #507. @olleolleolle please give a review. 😃


olleolleolle:
Thanks! Is it possible to figure out a test for this?

lib/github_changelog_generator/generator/generator_generation.rb
```diff
@@ -21,7 +21,8 @@ def compound_changelog
 
       log += File.read(options[:base]) if File.file?(options[:base])
 
-      log += "\n\n\\* *This Change Log was automatically generated by [github_changelog_generator](https://github.com/skywinder/Github-Changelog-Generator)*"
+      credit_line = "\* *This Change Log was automatically generated by [github_changelog_generator](https://github.com/skywinder/Github-Changelog-Generator)*"
+      log += "\n\n\#{credit_line}" if log.include? credit_line == false
```
Lucashuang0802: @olleolleolle do you know where the extra '\'? Should it be `"\n\n#{credit_line}"`?

Lucashuang0802:
@olleolleolle perhaps we can check there is only one `credit_line` text the whole file. WDYT?

olleolleolle:
For an integration test approach:

- Generate a file CHANGELOG.md (or have a fixture file) which has 1 credit_line in it
- Re-generate with some data (not important what data, just enough so that the credit_line code will be used)
- Expect that only 1 credit line is in the resulting file

Then, as a follow-up (next) feature:


- Generate a file CHANGELOG.md (or have a fixture file) which has _2 credit_lines_ in it
- Re-generate with some data (not important what data, just enough so that the credit_line code will be used)
- Expect that only 1 credit line is in the resulting file

Issue #541, opened 2017-06-28
Credit line bug


Lord-Haji:
https://github.com/Necrobot-Private/NecroBot/commit/4ba9964d4008a63dad2290fd06c5cac760d54fd5#diff-4ac32a78649ca5bdd8e0ba38b7006a1e
```diff
-\* *This Change Log was automatically generated by [github_changelog_generator](https://github.com/skywinder/Github-Changelog-Generator)*
+#{credit_line}
```

Lucashuang0802:
@Lord-Haji @olleolleolle sorry for this buggy change. I will take a look. 😄

Lucashuang0802:
@Lord-Haji thanks for reporting. The change is in.

Lord-Haji:
@Lucashuang0802 Thanks!

PR #542, opened 2017-06-28
Update generator_generation.rb

Lucashuang0802:
Fix the syntax so `credit_line` is recognizable as a variable. #541

I have tested it and worked well. @olleolleolle


olleolleolle:
Maybe you can author a test-case for this change?

Lucashuang0802:
@olleolleolle yes, it is WIP. 😄

PR #501, opened 2017-05-15
Remove --between-tags option

Lucashuang0802:

PR #550, opened 2017-07-27
Fix section mapping, hiding untagged PRs, and hiding untagged issues

hunner:
This is a rebase and resubmission of #453 to pass rubocop/tests and remove `--between-tags` references because it was removed in #501


olleolleolle:
I am grateful for this reasoned careful change.

PR #551, opened 2017-08-09
warn_if_nonmatching_regex should display proper help message when used with exclude-tags-regex

lacostej:
Problem: if regexp is taken from `exclude-tags-regex` and matches no tag, the warning message points to `exclude-tags`.

### DIFF
--- a/lib/github_changelog_generator/generator/generator_tags.rb
+++ b/lib/github_changelog_generator/generator/generator_tags.rb
@@ -177,18 +177,19 @@ module GitHubChangelogGenerator
 
     def apply_exclude_tags(all_tags)
       if options[:exclude_tags].is_a?(Regexp)
-        filter_tags_with_regex(all_tags, options[:exclude_tags])
+        filter_tags_with_regex(all_tags, options[:exclude_tags], "--exclude-tags")
       else
         filter_exact_tags(all_tags)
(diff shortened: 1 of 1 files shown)


lib/github_changelog_generator/generator/generator_tags.rb
```diff
@@ -177,18 +177,19 @@ def filter_excluded_tags(all_tags)
...
     def apply_exclude_tags_regex(all_tags)
-      filter_tags_with_regex(all_tags, Regexp.new(options[:exclude_tags_regex]))
+      regex = Regexp.new(options[:exclude_tags_regex])
+      filter_tags_with_regex(all_tags, regex, "--exclude-tags-regex")
     end
 
-    def filter_tags_with_regex(all_tags, regex)
-      warn_if_nonmatching_regex(all_tags)
+    def filter_tags_with_regex(all_tags, regex, regex_option_name)
+      warn_if_nonmatching_regex(all_tags, regex, regex_option_name)
```
olleolleolle: Cool change.

How do you feel about the parameter name `pattern` vs `regex`?

olleolleolle: Don't worry about it, we'll keep what was there.

PR #553, opened 2017-08-22
Update license date

benhc123:
to 2017

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,5 @@
 The MIT License (MIT)
-Copyright (c) 2016 Petr Korolev
+Copyright (c) 2017 Petr Korolev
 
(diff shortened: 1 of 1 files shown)


benhc123:
Surely this is one possible exception to this rule. It's just a date!

olleolleolle:
Hi! Thanks for contributing to GitHub Changelog Generator!

Licenses and copyright notices are a tricky landscape.

Please read this SO thread: [The copyright is not the same as the license.](https://softwareengineering.stackexchange.com/questions/210472/is-renewal-of-mit-license-needed-on-github-at-the-beginning-of-each-year)

With that thread as background, you can compare [an example of a project where they consider adding another year to the copyright notice](https://github.com/google/material-design-lite/issues/3523)

So, if we wish to improve the copyright notice, it _could_ be extended using a range of years in which changes were made to the software. Or, we could just leave it at the starting point.

Do you wish to make the change to add 2017 to the list of years in which changes have been made? Or turn it into a range?

olleolleolle:
Thanks for bringing this up – I merged your change with a variation: we now use a range of years.

Issue #546, opened 2017-07-16
Markdown: add a newline before the Full Changelog links


olleolleolle:
Formatting suggestion: add a newline between "## <version>" and the "Full Changelog" links -- some markdown parsers don't like things immediately after headers.

ianroberts131:
Ok if I take a stab at this? Looking to get started with open source development.

olleolleolle:
@ianroberts131 Go ahead!

ianroberts131:
I \*think\* I've located where the change should be made, so decided to try to throw a string there and test to see if I was on the right track...but I'm having trouble figuring out how to test it.

I believe the change needs to be made in `generator_generation.rb`, Line 69.
Here was how I planned to test this:
```ruby
log += if newer_tag_name.equal?(options[:unreleased_label])
         "## [#{newer_tag_name}](#{release_url})\n\n"
       else
         "## [#{newer_tag_name}](#{release_url}) (#{time_string})\n HELLLOOOOOOOO"
       end
```
I think I just need to insert another `\n`, but wasn't sure this would be visible to me so figured I'd try something that would be very obvious.

You can see the code here: https://github.com/ianroberts131/github-changelog-generator/blob/add-newline-after-version/lib/github_changelog_generator/generator/generator_generation.rb

In order to test the code, I've tried a couple of things:
1. Fork the changlog_test repository and create a Gemfile that references my forked version of github-changelog-generator
2. Create my own test repository, and then run the `github_changelog_generator` command locally.

Each time I run the command, I get the following output, but nothing happens:
```ruby
C:/Ruby23-x64/lib/ruby/2.3.0/logger.rb:711:in `initialize': No such file or directory @ rb_sysopen - /tmp/github-changelog-logger.log (Errno::ENOENT)
        from C:/Ruby23-x64/lib/ruby/2.3.0/logger.rb:711:in `open'
        from C:/Ruby23-x64/lib/ruby/2.3.0/logger.rb:711:in `create_logfile'
        from C:/Ruby23-x64/lib/ruby/2.3.0/logger.rb:705:in `rescue in open_logfile'
        from C:/Ruby23-x64/lib/ruby/2.3.0/logger.rb:702:in `open_logfile'
        from C:/Ruby23-x64/lib/ruby/2.3.0/logger.rb:695:in `set_dev'
        from C:/Ruby23-x64/lib/ruby/2.3.0/logger.rb:635:in `initialize'
[...]

PR #556, opened 2017-08-28
Add HR breaks for better readability

3minus1:
[First Timer] Another enhancement similar to Issue #546


olleolleolle:
Thanks for making this effort.

Adding markup for readability is not the way I want us to take. The display of HTML markup can be styled using CSS.

However!

Being more able to control output _is_ something I do care about, though. Your needs are important – even if I won't add `<hr>` tags to _my_ changelog, you may want to adorn yours with it.

I'm closing this PR, but I want there to be a way to design/decide more about the output than what we have currently ("one size fits all"). I will file an Issue about this.

Thanks again, for taking the time to contribute to GitHub Changelog Generator.

PR #566, opened 2017-10-04
filter tags correctly when `since_tag` is set to most recent tag

### LINKED ISSUE #304: since_tag seems to not be working

### LINKED ISSUE #555: since_tag doesn't work when tag specified is the latest tag
If `since_tag` is the latest tag available, the `unreleased` section contains *all* PRs/issues (not just the ones since the date of the since_tag).

Crunch09:
Hey,
before this fix [`filtered_tags`](https://github.com/skywinder/github-changelog-generator/blob/master/lib/github_changelog_generator/generator/generator_generation.rb#L132) was empty which caused a fallback to the last tag in the sorted_tags array (which is the oldest tag).
There for no issues and PRs were filtered in that case.

fixes #555
fixes #304

### DIFF
--- a/lib/github_changelog_generator/generator/generator_tags.rb
+++ b/lib/github_changelog_generator/generator/generator_tags.rb
@@ -129,7 +129,7 @@ module GitHubChangelogGenerator
       if tag
         if all_tags.map { |t| t["name"] }.include? tag
           idx = all_tags.index { |t| t["name"] == tag }
-          filtered_tags = if idx.present?
+          filtered_tags = if idx
                             all_tags[0..idx]
                           else
                             []
--- a/lib/github_changelog_generator/generator/generator_tags.rb
+++ b/lib/github_changelog_generator/generator/generator_tags.rb
@@ -129,7 +129,7 @@ module GitHubChangelogGenerator
       if tag
         if all_tags.map { |t| t["name"] }.include? tag
           idx = all_tags.index { |t| t["name"] == tag }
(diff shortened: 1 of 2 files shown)


Crunch09:
Exactly! Thanks for the thorough explanation! 😅 
> That means that if idx.present? could become if idx.

I'll change this, wasn't sure which way you prefer.

olleolleolle:
@Crunch09 Florian, **thank you** so much for taking up this issue and building this fix.

Crunch09:
@olleolleolle You're very welcome! Thank you for maintaining this gem! Great project 🙇

PR #574, opened 2017-10-13
Add option --require to load custom Ruby code before starting

olleolleolle:
This PR adds a feature and an option `--require`: it takes a comma-separated list of paths to custom Ruby files to `require`. This happens just before starting the changelog generation.

Example:

```
github_changelog_generator --require ./.changelog_hacks.rb,./.changelog_hacks2.rb
```

- [x] documentation of the new option

### DIFF
--- a/lib/github_changelog_generator/generator/generator_generation.rb
+++ b/lib/github_changelog_generator/generator/generator_generation.rb
@@ -6,6 +6,7 @@ module GitHubChangelogGenerator
     #
     # @return [String] Generated change log file
     def compound_changelog
+      options.load_custom_ruby_files
       fetch_and_filter_tags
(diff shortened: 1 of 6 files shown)


olleolleolle:
I was inspired to add this because of the Pivotal Tracker question. In order to make things more custom for them, they need a place to put code.

Perhaps a new default-required filename could be the default list, to be required if found.

olleolleolle:
#575 was the missing meta-documentation "How do I regenerate man page documentation?"

olleolleolle:
```
    -e, --load FILE                  A Ruby script to load before running command.
        --plugin PLUGIN              Load a YARD plugin (gem with `yard-' prefix)
```

Issue #577, opened 2017-10-15
bug: Output on aborting when unknown user and project only prints out banner, not why it aborted


Furtif:
Use `bundle info [gemname]` to see where a bundled gem is installed.
bundle exec github_changelog_generator
Usage: github_changelog_generator [options]
Command exited with code 1

olleolleolle:
I’m not quite following. 

Can you explain this with words like “I’m trying to achieve ___” or “When I do ___ this happens ___ but I expected ___.”

Furtif:
appveyor bug 
```
Switched to a new branch 'Appveyor'
SET PATH=C:\Ruby23-x64\bin;%PATH%
ruby --version
ruby 2.3.3p222 (2016-11-21 revision 56859) [x64-mingw32]
where ruby
C:\Ruby23-x64\bin\ruby.exe
C:\Ruby193\bin\ruby.exe
gem update --system --no-document --no-post-install-message
RubyGems 2.6.14 installed
=== 2.6.14 / 2017-10-09
Security fixes:
* Whitelist classes and symbols that are in loaded YAML.
  See CVE-2017-0903 for full details.
  Fix by Aaron Patterson.
=== 2.6.13 / 2017-08-27
Security fixes:
* Fix a DNS request hijacking vulnerability.
  Fix by Samuel Giddins.
* Fix an ANSI escape sequence vulnerability.
  Fix by Evan Phoenix.
* Fix a DOS vulernerability in the `query` command.
  Fix by Samuel Giddins.
* Fix a vulnerability in the gem installer that allowed
  a malicious gem to overwrite arbitrary files.
  Fix by Samuel Giddins.
=== 2.6.12 / 2017-04-30
Bug fixes:
* Fix test_self_find_files_with_gemfile to sort expected files. Pull
  request #1880 by Kazuaki Matsuo.
* Fix issue for MinGW / MSYS2 builds and testing. Pull request #1879 by
  MSP-Greg.
* Fix gem open to open highest version number rather than lowest. Pull
  request #1877 by Tim Pope.
* Add a test for requiring a default spec as installed by the ruby
  installer. Pull request #1899 by Samuel Giddins.
* Fix broken --exact parameter to gem command. Pull request #1873 by Jason
  Frey.
* [Installer] Generate backwards-compatible binstubs. Pull request #1904
  by Samuel Giddins.
* Fix pre-existing source recognition on add action. Pull request #1883 by
  Jonathan Claudius.
[...]

PR #578, opened 2017-10-15
Aborting on missing --user and --project prints all of usage

olleolleolle:
Thanks to investigations by @Furtif, this PR adds clarity to the abort message which happens if there's no user or project information after options and configurations have been parsed.

Before this change, the abort would print out only the name of the program and the "usage", which was incredibly bare.

- [ ] Explain to the user why their invocation did not work: "Tell us which user and project to work on. Options --user and --project, or settings to that effect."

See #577


olleolleolle:
@Furtif Thanks for your patience.

Furtif:
no worries 👍

PR #579, opened 2017-10-20
Make 'change log' a single word

mesaugat:
This PR basically replaces `change log` with `changelog` across all files.

There were inconsistencies in various files regarding the use of `change log` and `changelog`. I think `changelog` is the correct way to write it.


spec/unit/octo_fetcher_spec.rb
```diff
@@ -265,7 +265,7 @@
                             "diff_url"  => "https://github.com/skywinder/changelog_test/pull/21.diff",
                             "patch_url" => "https://github.com/skywinder/changelog_test/pull/21.patch" },
                         "body"           =>
-                          "to test https://github.com/skywinder/github-changelog-generator/pull/305\r\nshould appear in change log with #20" }
+                          "to test https://github.com/skywinder/github-changelog-generator/pull/305\r\nshould appear in changelog with #20" }
```
olleolleolle: Undo the changes in this file.

They refer to data in another repository, the changing of which is outside the scope of this find-and-replace PR.

spec/unit/octo_fetcher_spec.rb
```diff
@@ -265,7 +265,7 @@
                             "diff_url"  => "https://github.com/skywinder/changelog_test/pull/21.diff",
                             "patch_url" => "https://github.com/skywinder/changelog_test/pull/21.patch" },
                         "body"           =>
-                          "to test https://github.com/skywinder/github-changelog-generator/pull/305\r\nshould appear in changelog with #20" }
+                          "to test https://github.com/skywinder/github-changelog-generator/pull/305\r\nshould appear in change log with #20" }
```
olleolleolle: The "title" (in this document, too) 

```
"title" => "Merged br (should appear in changelog with  #20)",
```

needs to go back to what it was before.

mesaugat: Yes, just saw this.

PR #587, opened 2017-11-07
Add option --configure-sections, --add-sections, --include-merged

eputnam:
There's a lot in this PR.
- Added a Section class to more easily make the other changes and
  hopefully add flexibility for the future
- Added an option called `configure_sections` that allows you create
  your own custom sections. It blows away all other sections and uses
only the ones you give it.
- Added an option called `add_sections` that allows you to add_sections
  to the default section set
- Added an option called `include_merged` that can be used when
  configure_sections is defined. Configure sections blows away any and
all default sections so to get this one back, you have to set this
option.
- Added tests for this stuff

@HAIL9000 was a co-author. Because of a little git snafu, I accidentally
squashed all of our work into one so it looks like it was just me.


HAIL9000:
This looks so good! Thank you for running with it. Should we add our new options to the [docs](https://github.com/skywinder/github-changelog-generator/blob/master/man/git-generate-changelog.md)...?

olleolleolle:
This is a big PR. It may take a few turns of code review to get into the codebase, but don't be discouraged by that.

@HAIL9000 is right, there are a few places where all configuration options are to be placed. That is, duplicated in different formats.

olleolleolle:
There are a few style details to polish (later) in this code. RuboCop has its set of complaints.

If we can get more use out of the class `Section`, I'd be happy. It seems a very good addition to the codebase. Just looking quickly, perhaps a query method `Section#merged?` would help in some places?

Could the presence of the class `Section` signal the need for a `Release`, as well?

eputnam:
@olleolleolle can you elaborate a little on both of those? i'm not sure i see the need for the merged? function or how the signaling for a release would work.

hunner:
I have fixed the rubocop failures and reorganized code relating to generating specific releases and sections of each release out into their own classes outside of the main generation code. The methods otherwise remain mostly the same (a few naming tweaks because "section" and "log" were overloaded terms).

This is still a Work In Progress as I would like to add more tests.

lib/github_changelog_generator/generator/entry.rb
```diff
@@ -0,0 +1,309 @@
...
+
+    # Generates log entry with header and body
+    #
+    # @param [Array] pull_requests List or PR's in new section
+    # @param [Array] issues List of issues in new section
+    # @param [String] newer_tag Name of the newer tag. Could be nil for `Unreleased` section
+    # @param [Hash, nil] older_tag Older tag, used for the links. Could be nil for last tag.
+    # @return [String] Ready and parsed section
+    def create_entry_for_tag(pull_requests, issues, newer_tag, older_tag = nil)
+      newer_tag_link, newer_tag_name, newer_tag_time = detect_link_tag_time(newer_tag)
```
hunner: Should an entry be told what the tags link/name/time is, or should it be able to figure it out just given a tag?

If the former, this logic moves up to the generator. If the latter, the methods for tag munging need to move into the entry class.

lib/github_changelog_generator/generator/entry.rb
```diff
@@ -0,0 +1,309 @@
...
+
+      github_site = @options[:github_site] || "https://github.com"
+      project_url = "#{github_site}/#{@options[:user]}/#{@options[:project]}"
+
+      # If the older tag is nil, go back in time from the latest tag and find
+      # the SHA for the first commit.
+      older_tag_name =
+        if older_tag.nil?
+          @fetcher.commits_before(newer_tag_time).last["sha"]
```
hunner: Should an entry fetch it's preceding commits, or is this logic part of the larger changelog generation?

If the former, then the entry class needs access to the fetcher (pass through options?). If the latter, then this needs to move out of the entry class back to the generator class.

lib/github_changelog_generator/generator/entry.rb
```diff
@@ -0,0 +1,309 @@
...
+
+      set_sections_and_maps
+
+      @content = generate_header(newer_tag_name, newer_tag_link, newer_tag_time, older_tag_name, project_url)
+
+      @content += generate_body(pull_requests, issues)
+
+      @content
+    end
+
```
hunner: All following methods should be private, and `create_entry_for_tag` should be tested.

lib/github_changelog_generator/generator/entry.rb
hunner: The latter would be the better design decision, but `get_time_of_tag` is called by `detect_link_tag_time` and used in multiple other places, so moving `detect_link_tag_time` to the `Entry` class would cause issues around that.

Separating the tag utility methods out into their own `GeneratorTags` module instead of being part of either the `Generator` or `Entry` classes would be preferable, but is beyond the scope of this PR.

lib/github_changelog_generator/generator/entry.rb
hunner: The `create_entry_for_tag` method should not make calls out to github, so instead of having the fetcher passed through in the case of a nil older tag, the older tag name should be a required parameter and the generator class should either pass the older tag name or the first commit sha in the absence of an older tag.

The `older_tag_link` variable used in `generate_header` is misnamed since it is actually just a tag link, not a name. It is also possibly a sha and not a tag, but I'll ignore that for now.

lib/github_changelog_generator/generator/entry.rb
hunner: The `newer_tag` values of name/link/time should be consistently handled in that order inside the `Entry` class. I have done some small refactoring to make this so.

lib/github_changelog_generator/generator/section.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    # @param [Array] issues List of issues on sub-section
+    # @param [String] prefix Name of sub-section
+    # @return [String] Generate section content
+    def generate_content(issues = @issues, prefix = @prefix)
+      content = ""
+
+      if issues.any?
+        content += "#{prefix}\n\n" unless @options[:simple_list]
```
hunner: The `@options` hash is scattered around the various classes and passed around as needed but never modified once parsed. It would be nice if the options were centrally accessible instead of passing them around, but that is out of scope for this PR.

Instead, I'll just make `simple_list` an invocation option to be removed later.

hunner: Eh, there are several more options needed by the `Section` class. I guess I'll pass `@options` through and it can be refactored in another PR.

hunner:
@olleolleolle Green and ready for your review again \o/.

eputnam:
tagging @kvolkovich-sc since you seem to have some interest in the feature. feel free to have a look at what we've put together!

skywinder:
Very powerful PR. Thank you all for your efforts, @eputnam @hunner 🎉

lib/github_changelog_generator/generator/entry.rb
```diff
@@ -0,0 +1,256 @@
...
+    #
+    # @return [Hash] map of 'section name' => section object
+    def section_map
+      map = {}
+
+      @sections.each do |section|
+        map[section.name] = section
+      end
+
+      map
```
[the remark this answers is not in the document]
hunner: Good catch!

lib/github_changelog_generator/generator/entry.rb
```diff
@@ -0,0 +1,256 @@
...
+        header += "[Full Changelog](#{project_url}/compare/#{older_tag_name}...#{newer_tag_link})\n\n"
+      end
+
+      header
+    end
+
+    # Generates complete body text for a tag (without a header)
+    #
+    # @param [Array] issues
+    # @param [Array] pull_requests
```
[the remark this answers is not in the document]
hunner: Okay. I updated the methods to take the order `pull_requests, issues` as well for consistency.

lib/github_changelog_generator/generator/entry.rb
```diff
@@ -0,0 +1,232 @@
...
+
+    # Generates complete body text for a tag (without a header)
+    #
+    # @param [Array] pull_requests
+    # @param [Array] issues
+    # @returns [String] ready-to-go tag body
+    def generate_body(pull_requests, issues)
+      body = ""
+      body += main_sections_to_log(pull_requests, issues)
+      body += merged_section_to_log(pull_requests) if (@options[:pulls] && @options[:add_pr_wo_labels]) || (@options.configure_sections? && @options[:include_merged])
```
hunner: It looks like the logic here is not correct. `:include_merged` defaults to true, but seems like it should default to false. Additionally, the `:configure_section` option does not override the `:add_pr_wo_labels` option as it appears to be meant to do.

The UX here needs a little thinking through.

hunner:
Okay, it looks like the `add_pr_wo_labels` option applies to normal generation and custom section generation now. No `include_merged` option needed.

eputnam:
ready for you again, @olleolleolle @skywinder

lib/github_changelog_generator/parser_file.rb
```diff
@@ -96,7 +96,9 @@ def convert_value(value, option_name)
       pr_wo_labels: :add_pr_wo_labels,
       pull_requests: :pulls,
       filter_by_milestone: :filter_issues_by_milestone,
-      github_api: :github_endpoint
+      github_api: :github_endpoint,
+      configure_sections: :configure_sections,
+      add_sections: :add_sections
```
olleolleolle: These are only needed here if they aren't the same - as you can see, all the other pairs are about irregular names.

So, do not add the new option names here.

spec/unit/generator/entry_spec.rb
```diff
@@ -0,0 +1,351 @@
+# frozen_string_literal: true
+
+describe GitHubChangelogGenerator::Entry do
```
olleolleolle: Please use the form `RSpec.describe` (not `describe`) for the outermost describe.

spec/unit/generator/entry_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+        **Merged pull requests:**
+
+        - pr no labels [\\#10](https://github.com/owner/repo/pull/10) ([user1](https://github.com/user1))
+
+CHANGELOG
+                                                                                                                         )
+    end
+  end
+  describe "#parse_sections" do
+    before :each do
```
olleolleolle: `:each` is deprecated RSpec syntax. 

That setting is today called `:example`, but it is also the default.

So you can say `before do ... end`  instead.

spec/unit/generator/entry_spec.rb
olleolleolle: You can structure tests inside a module, like this:

```ruby
  RSpec.describe Entry do
    # ...in here, names will be in the namespace of the module.
    # so that you can use shorter names.
  end
end

File: spec/unit/generator/entry_spec.rb [hunk 11]
```diff
@@ -0,0 +1,351 @@
...
+        require "json"
+
+        sections_json = JSON.parse(sections_string)
+        sections_array.each_index do |i|
+          expect(sections_array[i].name).to eq sections_json.first[0]
+          expect(sections_array[i].prefix).to eq sections_json.first[1]["prefix"]
+          expect(sections_array[i].labels).to eq sections_json.first[1]["labels"]
+          expect(sections_array[i].issues).to eq []
+          sections_json.shift
+        end
```
olleolleolle: This place is a candidate for using [aggregate_failures,](https://relishapp.com/rspec/rspec-core/docs/expectation-framework-integration/aggregating-failures) the RSpec feature.

There are _many_ `expect`s in the same `it` here, and if there's a failure, things stop executing.

Look that feature up, it's quite cool.

spec/unit/generator/entry_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+      end
+
+      it "returns correctly constructed sections" do
+        sections_array.each_index do |i|
+          expect(sections_array[i].name).to eq sections_hash.first[0].to_s
+          expect(sections_array[i].prefix).to eq sections_hash.first[1][:prefix]
+          expect(sections_array[i].labels).to eq sections_hash.first[1][:labels]
+          expect(sections_array[i].issues).to eq []
+          sections_hash.shift
+        end
```
olleolleolle: Right, here's another `it` with many expects.

spec/unit/generator/entry_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+      "pull_request" => true,
+      "title" => "pr #{title}",
+      "labels" => labels.map { |l| label(l) },
+      "number" => number,
+      "html_url" => "https://github.com/owner/repo/pull/#{number}",
+      "user" => user.merge("html_url" => "https://github.com/#{user['login']}")
+    }
+  end
+
+  def get_titles(issues)
```
olleolleolle: Could we change this name to be more about the result of the call?

Example method name:

- `titles_for(issues)`

spec/unit/generator/entry_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+      it "assigns issues to the correct sections" do
+        foo_section = @arr.select { |section| section.name == "foo" }[0]
+        enhancement_section = @arr.select { |section| section.name == "enhancements" }[0]
+        bug_section = @arr.select { |section| section.name == "bugs" }[0]
+
+        expect(get_titles(foo_section.issues)).to eq(["issue test1", "issue all the labels", "pr test1", "pr all the labels"])
+        expect(get_titles(enhancement_section.issues)).to eq(["pr enhance"])
+        expect(get_titles(bug_section.issues)).to eq(["issue bugaboo"])
+        expect(get_titles(pull_requests)).to eq(["pr no labels"])
+      end
```
olleolleolle: Many `expect`s here, too.

spec/unit/generator/entry_spec.rb
```diff
@@ -98,253 +100,262 @@ def default_sections
...
+      before do
+        subject { described_class.new }
       end
+      context "valid json" do
+        let(:sections_string) { "{ \"foo\": { \"prefix\": \"foofix\", \"labels\": [\"test1\", \"test2\"]}, \"bar\": { \"prefix\": \"barfix\", \"labels\": [\"test3\", \"test4\"]}}" }
+
+        let(:sections_array) do
+          [
+            GitHubChangelogGenerator::Section.new(name: "foo", prefix: "foofix", labels: %w[test1 test2]),
+            GitHubChangelogGenerator::Section.new(name: "bar", prefix: "barfix", labels: %w[test3 test4])
```
olleolleolle: Now, it ought to be possible to just say "Section.new(...)" here.

lib/github_changelog_generator/generator/entry.rb
```diff
@@ -0,0 +1,232 @@
+require "github_changelog_generator/generator/section"
+
+module GitHubChangelogGenerator
+  class Entry
+    attr_reader :content
+
+    def initialize(options = {})
+      @content = ""
+      @options = Options.new(options)
```
olleolleolle: Could we decide that it must be an `Options` that is passed in here?

hunner:
Rebased to resolve conflicts. Anything else you'd like to see?

olleolleolle:
@hunner One small thing! 

We really could have some doc blocks on these new classes, with more-descriptive words, such as this great section:

> I reasoned that a changelog is a series of "entries" of all tagged
> releases plus an extra entry for the unreleased entry. Each entry is
> comprised of a header and a series of "sections" for that entry. Each
> section is comprized of a list of issues and/or pull requests for that
> entry. So the log contains entries, entries contain sections, and
> sections contain issues & prs. I have structured the classes around this idea.

lib/github_changelog_generator/generator/entry.rb
```diff
@@ -0,0 +1,232 @@
...
+    # @return [Array] array of Section objects
+    def parse_sections(sections_desc)
+      require "json"
+
+      sections_desc = sections_desc.to_json if sections_desc.class == Hash
+
+      begin
+        sections_json = JSON.parse(sections_desc)
+      rescue JSON::ParserError => e
+        raise "There was a problem parsing your JSON string for secions: #{e}"
```
olleolleolle: Typo: `secions` => `sections`

lib/github_changelog_generator/generator/entry.rb
```diff
@@ -0,0 +1,232 @@
...
+      end
+
+      sections_arr = []
+
+      sections_json.each do |name, v|
+        sections_arr << Section.new(name: name.to_s, prefix: v["prefix"], labels: v["labels"], options: @options)
+      end
+
+      sections_arr
```
olleolleolle: We can leave off naming a local variable here, and only return the result of `#map`

```ruby
sections_json.map do |name, v|
  Section.new(name: name.to_s, prefix: v["prefix"],
              labels: v["labels"], options: @options)
end 
```

olleolleolle:
It's been a marathon, all!

👍

hunner:
Thank you for working with us to get this merged too; and thank you for such commitment to maintaining this useful tool!

Issue #591, opened 2017-11-23
Dockerfile and Docker automated builds


ferrarimarco:
Hi @olleolleolle 

Would you mind accepting a pull request for a Dockerfile? I can also configure a Docker Hub automated build that runs on each tag (even on each commit if that is the desired behaviour).

olleolleolle:
I’d be happy to have a PR!

ferrarimarco:
Ok, I'm working on it.

How do you publish new releases on Rubygems? I need this info to think about the Dockerfile and the configuration of automated builds

ferrarimarco:
@olleolleolle here it is: #592

olleolleolle:
@ferrarimarco I use the `rake release` Rake task to make new versions?

ferrarimarco:
As soon as (and if) #592 is merged I'll configure the Docker Hub automated build if it's fine for you

ferrarimarco:
Ok I can now configure the Docker automated build. Which prefix should we use for the image?

ferrarimarco:
@olleolleolle I don't have access to this repo configuration, so I cannot enable the automatic build all by myself.

Let's explore the options:

### Automated build with Docker hub

If we want to use (like previously suggested) Docker Hub for builds we should (since skywinder is a user account on GitHub and not GitHub Organization):
1. Register a new account on Docker Hub (let's call it docker-hub-user in this example)
1. Link docker-hub-user with github.com/skywinder (anyone doing this must know github.com/skywinder credentials)
1. Let me add docker-hub-user in the skywinder Docker Hub organization (I created it moments ago)
1. Use docker-hub-user to create a new automated build from GitHub and choose "skywinder" as namespace when creating it. Since docker-hub-user has been linked to github.com/skywinder, it will have access to this repo
1. Let me configure the automated build for you (since I'm a member of the skywinder organization I have the rights to do so)

### Automated build with Travis CI

Each time a tag is pushed, we use Travis CI to build the Docker image and push it under the skywinder Docker Hub organization. For this we need update the `.travis.yml` (I can do that) but you (or someone who has the proper rights) should update the Travis build configuration because we need to set the Docker Hub credentials in Travis CI.

This approach has the added benefit that we can use Travis CI environment variables (in our case we are going to use the git tag) to tag the docker image and to pass it as a build time variable instead of hardcoding the version in the Dockerfile.

[...]

PR #592, opened 2017-11-23
Implemented a Dockerfile

ferrarimarco:
In this pull request I implemented a Dockerfile and added a couple of notes to the README (to refine, asking for help on this).

Relevant issue: #591 

# Implementation

## Base Image

I chose `ruby:2.4-alpine3.6` as a base for this Docker image after checking CHANGELOG.md (see #487).
Such base always points to the latest 2.4.x release. We can explicitely point to `ruby:2.4.1-alpine3.6` if desired (but then we are creating another bit to update on each Ruby version bump).

I picked up the latest Alpine variant to reduce the size of the resulting image.

## GitHub Changelog Generator Installation

The installation is carried out by the following command:

```dockerfile
RUN gem install github_changelog_generator
```
This means that `gem` will install the current default version. We can extend this to allow the installation of pre-release versions and even by explicitely specify the version to install (this means that the Dockerfile has to be (maybe automatically) modified on each version bump). Let talk about this.

I can also configure Travis CI to build the Docker image on each commit, to test if everything is fine on this side. Let's decide first how to handle the version to install.

## Volumes

We declare `/usr/local/src/your-app` (feel free to find a better name for the directory) as a Docker Volume and then we expect that the user bind mounts his repository root to `/usr/local/src/your-app`, as we suggest in the README.

# README

As stated, I'm asking for help here. Can you please check if the only addition is enough?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM ruby:2.4-alpine3.6
+
+LABEL maintainer "olle.jonsson@gmail.com"
+
+RUN gem install github_changelog_generator
+
+ENV SRC_PATH /usr/local/src/your-app
+RUN mkdir -p $SRC_PATH
+
+VOLUME [ "$SRC_PATH" ]
+WORKDIR $SRC_PATH
+
+CMD ["--help"]
+ENTRYPOINT ["github_changelog_generator"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,9 @@ FROM ruby:2.4-alpine3.6
 
 LABEL maintainer "ferrari.marco@gmail.com"
 
-RUN gem install github_changelog_generator
+ENV GITHUB_CHANGELOG_GENERATOR_VERSION "1.14.3"
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM ruby:2.4-alpine3.6
+
+LABEL maintainer "olle.jonsson@gmail.com"
```
olleolleolle: Can you be the maintainer of this Dockerfile? Care to join?

olleolleolle:
Auto-bumping the build version in a script seems like an OK release process.

I use [the `bump` gem](https://github.com/gregorym/bump), to increase the version number of the gem. That can be used to change the Dockerfile, too, afterwards. 

I'm thinking we can wrap this inside a Rake task as a step later.

Dockerfile
ferrarimarco: Sure! :+1:

ferrarimarco: Done

ferrarimarco:
@olleolleolle done, added the explicit version

PR #600, opened 2017-12-12
Full stop

benhc123:
A minor cosmetic correction


olleolleolle:
A thought: is that even the actual message?

benhc123:
Good question! I'll investigate.

benhc123:
It seems like the actual error might look like this
`API rate limit exceeded for xxx.xxx.xxx.xxx`

benhc123:
I think #601 looks better.

PR #602, opened 2017-12-13
Add $ or # to indicate whether a command needs to be run as root or n…

mueller-ma:
…ormal user


README.md
```diff
@@ -83,19 +83,23 @@ See also Troubleshooting.
...
-                                 --github-api="https://github.yoursite.com/api/v3/"
+       $ github_changelog_generator --github-site="https://github.yoursite.com" \
+                                  --github-api="https://github.yoursite.com/api/v3/"
 
 This generates a `CHANGELOG.md`, with pretty Markdown formatting.
 
 ### Params
 
-Type `github_changelog_generator --help` for details.
+Type `$ github_changelog_generator --help` for details.
```
olleolleolle: Here, to avoid confusing readers use an indented section.

```
Print help for all command-line options to learn more details:

    $ github_changelog_generator --help
```

README.md
```diff
@@ -43,20 +43,20 @@ release (or version) of the project."
 
 ## Installation
 
-    gem install github_changelog_generator
+    # gem install github_changelog_generator
```
olleolleolle: Here, I would like _not_ to say `$` nor `#`.

The way people install Ruby varies wildly. Their best way to use it is about how they installed it.

olleolleolle:
Thank you for improving the documentation!

I made some notes, and if you feel you have the energy, you can follow up on them, and if not, just say so, and I'll take it from there.

Thanks, again!

README.md
mueller-ma: I would say `#`. People who are unfamiliar with ruby have it installed via their package manager. People how are familiar with ruby, know that they dont need to run it as root.

Maybe `$ sudo` would be better here.

olleolleolle: Some people take the long way around and `$ [sudo] gem install ...`

mueller-ma: Thats a good compromise. Changed it.

README.md
```diff
@@ -83,19 +83,25 @@ See also Troubleshooting.
 
 -  Run this:
 
-      `github_changelog_generator -u github_username -p github_project`
-      `github_changelog_generator  github_username/github_project`
+	   $ github_changelog_generator -u github_username -p github_project
+	   
+	or
```
olleolleolle: Perhaps we can make a note here, about versions.

```
    or, on the 1.14.x (current stable release)
```

olleolleolle: this was what brought you all the way here, to fixing the documentation.

(Thanks for that!)

mueller-ma: Or just remove the old syntax? When people download the latest version, they dont need to know old stuff.

olleolleolle: The README describes things which are yet to be released.

The visitor may have installed the 1.14.x (stable), and not 1.15.x-pre. (pre-release).

Using `gem install ...` will install stable.

olleolleolle: So, they _may_ need to know this difference.

olleolleolle:
Thanks, those were all my notes! I'll merge this!

README.md
mueller-ma: Changed it.

PR #606, opened 2017-12-15
Update git-generate-changelog.html

benhc123:


olleolleolle:
I’m wondering, though, is this a regenerated result from the ronn file, or is it manually edited?

It might be that we still have a spelling error in the .ronn source file which creates this, which would overwrite this if used. 

Would you want to make the check for that, before I merge something which could be undone?

PR #618, opened 2018-01-27
Fix unlabeled, mixed labels, and unmapped labels handling

hunner:
- Issues/PRs filtered by --exclude-labels/--include-labels should never be added to sections.
- Issues/PRs with labels mapping to multiple sections should map to default sections order of breaking, then enhancement, then fix.
- Issues/PRs with labels mapping to multiple sections should map to default sections then the first-matching section in --add-sections.
- Issues/PRs with labels mapping to multiple sections should map to the first-matching section in --configure-sections.
- Issues/PRs with some mapped labels and some unmapped labels should respect the mapping rather than behaving like unmapped issues.
- Issues/PRs with only unmapped labels should be added to the --merge-prefix/--issue-prefix section.
- Issues/PRs with no labels should be added to the --merge-prefix/--issue-prefix section unless --add-pr-wo-labels/--add-issues-wo-labels is false.
- Filter both PRs and issues equally.

### DIFF
--- a/lib/github_changelog_generator/generator/entry.rb
+++ b/lib/github_changelog_generator/generator/entry.rb
@@ -25,26 +25,24 @@ module GitHubChangelogGenerator
     # @param [String] newer_tag_name Name of the newer tag. Could be nil for `Unreleased` section.
     # @param [String] newer_tag_link Name of the newer tag. Could be "HEAD" for `Unreleased` section.
     # @param [Time] newer_tag_time Time of the newer tag
-    # @param [Hash, nil] older_tag Older tag, used for the links. Could be nil for last tag.
-    # @return [String] Ready and parsed section
-    def create_entry_for_tag(pull_requests, issues, newer_tag_name, newer_tag_link, newer_tag_time, older_tag_name) # rubocop:disable Metrics/ParameterLists
+    # @param [Hash, nil] older_tag_name Older tag, used for the links. Could be nil for last tag.
+    # @return [String] Ready and parsed section content.
+    def generate_entry_for_tag(pull_requests, issues, newer_tag_name, newer_tag_link, newer_tag_time, older_tag_name) # rubocop:disable Metrics/ParameterLists
       github_site = @options[:github_site] || "https://github.com"
       project_url = "#{github_site}/#{@options[:user]}/#{@options[:project]}"
 
-      set_sections_and_maps
+      create_sections
 
       @content = generate_header(newer_tag_name, newer_tag_link, newer_tag_time, older_tag_name, project_url)
-
       @content += generate_body(pull_requests, issues)
-
       @content
     end
 
     private
 
-    # Creates section objects and the label and section maps needed for
-    # sorting
-    def set_sections_and_maps
+    # Creates section objects for this entry.
+    # @return [Nil]
+    def create_sections
       @sections = if @options.configure_sections?
                     parse_sections(@options[:configure_sections])
                   elsif @options.add_sections?
@@ -52,15 +50,14 @@ module GitHubChangelogGenerator
                   else
                     default_sections
                   end
-
-      @lmap = label_map
-      @smap = section_map
+      nil
     end
 
-    # Turns a string from the commandline into an array of Section objects
+    # Turns the argument from the commandline of --configure-sections or
+    # --add-sections into an array of Section objects.
     #
-    # @param [String, Hash] either string or hash describing sections
-    # @return [Array] array of Section objects
+    # @param [String, Hash] sections_desc Either string or hash describing sections
+    # @return [Array] Parsed section objects.
     def parse_sections(sections_desc)
       require "json"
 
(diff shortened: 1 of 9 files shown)


olleolleolle:
Good work! Fine analysis and useful in-code “why comments” in tricky places. 

Thank you for bringing this to the project!

I’ll watch this closely.

lib/github_changelog_generator/generator/entry.rb
```diff
@@ -26,41 +26,38 @@ def initialize(options = Options.new({}))
...
     # @param [Hash, nil] older_tag Older tag, used for the links. Could be nil for last tag.
-    # @return [String] Ready and parsed section
-    def create_entry_for_tag(pull_requests, issues, newer_tag_name, newer_tag_link, newer_tag_time, older_tag_name) # rubocop:disable Metrics/ParameterLists
+    # @return [String] Ready and parsed section content.
+    def generate_entry_for_tag(pull_requests, issues, newer_tag_name, newer_tag_link, newer_tag_time, older_tag_name) # rubocop:disable Metrics/ParameterLists
       github_site = @options[:github_site] || "https://github.com"
       project_url = "#{github_site}/#{@options[:user]}/#{@options[:project]}"
 
-      set_sections_and_maps
+      create_sections
```
eputnam: is it worth changing this method name to `generate_sections`? it would match better with everything around it.

lib/github_changelog_generator/generator/entry.rb
```diff
@@ -135,94 +112,92 @@ def generate_header(newer_tag_name, newer_tag_link, newer_tag_time, older_tag_na
...
     # @param [Array] issues
-    # @returns [String] ready-to-go tag body
+    # @returns [String] Content generated from sections of sorted issues & PRs.
     def generate_body(pull_requests, issues)
-      body = ""
-      body += main_sections_to_log(pull_requests, issues)
-      body += merged_section_to_log(pull_requests) if @options[:pulls] && @options[:add_pr_wo_labels]
-      body
+      sort_into_sections(pull_requests, issues)
+      @sections.map(&:generate_content).join
```
eputnam: i just learned what `Array.map(&:foo)` does. slick!

lib/github_changelog_generator/generator/entry.rb
olleolleolle: Reading the code, my personal take is “create_” reads easier. 

We could move to that naming instead? But to me, that’s not a merge blocker for this change. 

Thanks for keeping sharp eyes on This developing PR, @eputnam!

hunner: This method is about filling the `@sections` array with the `Section.new` objects to be filled out later on. I tried to make any `generate_*` methods be ones that return generated string content for the changelog.

hunner:
I think this is ready to go. It'd be nice if there were rubocop rules to catch missing `@param` and `@return` documentation.

olleolleolle:
There is a little gem called yard-junk you can run to check your work. It has that goal. 

The output it makes is more helpful in finding wrong docblocks, though. Perhaps not the exact tool you asked for.

hunner:
Thanks! I'll check it out.

olleolleolle:
@hunner If there are any YARD improvements to be made, they're not a blocker to merging, if you feel this is ready for merge.

hunner:
I added a few yard fixes pointed out by yard-junk just now, though not all of them were from these changes.

---

I'm not sure this is an enhancement. The command line arguments affected (`--add-pr-wo-labels`/`--add-issues-wo-labels`/`--configure-sections`/`--merge-prefix`/`--issue-prefix`/`--exclude-labels`/`--include-labels`) all existed before this change and had mostly-defined behavior, though some of the edge-cases were not defined or poorly handled. This PR is trying to bugfix around the edge cases of unlabeled/unmapped PRs and make the behavior more consistent with the various combinations of CLI arguments that were added in other recent PRs.

This change could be [compared to v1.14.3](https://github.com/skywinder/github-changelog-generator/compare/v1.14.3...hunner:unlabeled_unmapped?expand=1) since 1.15.0 hasn't been finalized yet. It seems that #530 (`--breaking-label`/`--breaking-labels`) & #587 (`--configure-sections`/`--add-sections`) are features and this PR rectifies the behavior between the new features and the pre-existing ones.

You could could call this a feature if the handling of unlabeled PRs and PRs with unmapped labels is the feature that this implements, perhaps? Though without adding any arguments to do so it's arguable that this behavior was just undefined and is now being defined (which is a bugfix).

hunner:
One change that I see this PR makes is that if `--include-labels` is specified, then only issues with those labels are included. Previous behavior was that issues with those labels plus unlabeled issues were included.

Since `--add-issues-wo-labels` and `--add-pr-wo-labels` default to true, perhaps those should control the unlabeled issue behavior and `--include-label` behavior should not affect unlabeled issue behavior.

olleolleolle:
I feel this is ready to merge to master. @hunner made a clear case for why. @hunner - any other opinion?

hunner:
I think this is good to go!

olleolleolle:
@hunner Thank you!

Issue #615, opened 2018-01-20
Specifying since-tag on a PR merge commit can cause that PR to be included


peppy:
When specifying a `since-tag` that corresponds to a PR merge commit (where the ref of the tag and the merge commit are equal) the PR is unexpectedly included in the changelog.

Tested on latest pre-release build using

`github_changelog_generator -u ppy -p osu-framework --since-tag changelog-20180120 --max-issues 500 --no-issues --release-branch master`

note that the state of this repository may change, let me know if i need to make an isolated test case.

hunner:
This may be fixed in master now.

Issue #580, opened 2017-10-23
Rebased PRs


AlexWayfer:
Hello!

In [AlexWayfer/flame](https://github.com/AlexWayfer/flame) I have [a list of commits](https://github.com/AlexWayfer/flame/commits/master?after=50b9806a3617384ad6b4cffe0651a9b53266ebbe+69) with [commit for version updating](https://github.com/AlexWayfer/flame/commit/cad968bd83548b1d1931a536954c32d363f9372c) and [commit with big refactoring](https://github.com/AlexWayfer/flame/commit/e34118f48d342fb8d9a11b6ab610c042dd2ff435). The second commit in list is after the first, but has earlier date-time (PR + `rebase`, I guess) . And the second commit appears in changelog for `4.18.1` version, but if fact it is for next release.

And one more thing: changelog has both (commit and issue), does it right?

Thank you.

Generated changelog:

```markdown
- Fix `flame` gem in template Gemfile [\#34](https://github.com/AlexWayfer/flame/pull/34) ([AlexWayfer](https://github.com/AlexWayfer))
- Add `Flame::Application.require\_dirs` method [\#33](https://github.com/AlexWayfer/flame/pull/33) ([AlexWayfer](https://github.com/AlexWayfer))
- Remove patch-versions of Ruby for Travis CI [\#32](https://github.com/AlexWayfer/flame/pull/32) ([AlexWayfer](https://github.com/AlexWayfer))

## [v4.18.1](https://github.com/AlexWayfer/flame/tree/v4.18.1) (2017-07-29)
[Full Changelog](https://github.com/AlexWayfer/flame/compare/v4.18.0...v4.18.1)

**Implemented enhancements:**

- Add `version` option for `Controller\#url\_to` method with static files [\#14](https://github.com/AlexWayfer/flame/issues/14)
- Raise error if no file for rendering [\#12](https://github.com/AlexWayfer/flame/issues/12)
- Don't allow change the order of optional parameters in path [\#7](https://github.com/AlexWayfer/flame/issues/7)
- Change routing system [\#3](https://github.com/AlexWayfer/flame/issues/3)
- Replace Array-based routes system with Hash-based [\#27](https://github.com/AlexWayfer/flame/pull/27) ([AlexWayfer](https://github.com/AlexWayfer))

**Fixed bugs:**

- Don't allow change the order of optional parameters in path [\#7](https://github.com/AlexWayfer/flame/issues/7)

[...]

Issue #560, opened 2017-09-14
Commit of release is seen as unreleased.


mjonker92:
We have created a new release from the latest master commit. The release notes for this release contain the commit of the last release.

Also after creating the release and running with `--unreleased-only` I still get the commit of the last release (which closed an issue) as the release notes.

pociej:
+1

hunner:
This may be fixed in master now.

Issue #497, opened 2017-05-11
Non-linear release histories not supported


codysoyland:
I may have missed a flag or something, but this doesn't appear to support release histories that include multiple maintained release branches.

As an example, see the changelog generated for github.com/pilosa/pilosa: https://gist.github.com/codysoyland/703101d5be09fc811cb3f7a6ab8d1e8e

As you can see, we made releases in the following order:

1. 0.2.4
2. 0.3.1
3. 0.2.5
4. 0.3.2

However, the changelog assumes that each of these releases was an evolution from the previous by order of date. But we are actually maintaining two release branches, a 0.2.x branch and a 0.3.x branch.

In other words, the previous version that is used to generate the change list and diff link should not be the previous tag by date, but the most recent tagged ancestor in git history.

Slamdunk:
Totally 👍 for this.

Maybe also related to https://github.com/skywinder/github-changelog-generator/issues/256

Supporting two or more branches is common, bugfixes go to the lowest supported branch and then merged in more recent branches, while latest branch keeps getting PR merged in the meantime.

This bug mixes the result that therefore must be manually cleared 😞

hunner:
#619 fixes this for PR merge detection but doesn't fix it for the generated diff links. Something like https://github.com/skywinder/github-changelog-generator/pull/619#issuecomment-370598725 would be needed.

mvz:
It is unclear to me what the desired result is. Should new changes in the 0.2.x series even appear in the 0.3.x changelog? If not, the `--release-branch` option introduced in #305 should help.

PR #619, opened 2018-02-10
Use git history to find PRs in a tag instead of time

### LINKED ISSUE #617: Some PRs logged for the wrong releases
1f68ef8 causes many PRs to be logged in the changelog incorrectly.

One problem caused by this this is because the `merged_at` date (on github's PR api) can be 1+ seconds later than the merge commit date because the PR is closed after the merge commit is pushed. This matters because if the merge commit of a PR is the tagged SHA then that PR is considered part of the next release (it's merged_at time is later than the date of the tagged SHA).

Another issue I haven't found the root cause yet is that some commits that are rebased may be mentioned as being in a later release, such as #274 and #275 ; #274 went out in 1.8.4 and #275 in 1.8.5, and all 1.15.x tags currently incorrectly place this in 1.8.5.

This should be fixed before 1.15.0 goes out. Otherwise changelogs generated with 1.14.x or earlier will be incorrectly reorganized.

hunner:
Fixes #617


hunner:
Currently the two extra round-trip calls to the github api per PR are agonizingly slow. This should be revised to use the existing git history instead of github's api.

What should GCG do for PRs that are "rebase and squash" PR merge strategies similar to how https://github.com/syl20bnr/spacemacs/ does it? I suspect that even using merged_at didn't work as no PRs are "merged" just closed and referenced in commit messages, so we can probably ignore that.

hunner:
Issues have no "merge" event so would have to continue to be time-based. I wonder what happens for issues that are opened and closed at points in the future?

zyphlar:
FYI I tried copy-pasting these file revisions into my local copy to try and fix things, and got two errors in succession:

`/var/lib/gems/2.3.0/gems/github_changelog_generator-1.14.3/lib/github_changelog_generator/generator/generator_fetcher.rb:49:in `sha_in_tag?': undefined local variable or method `tag' for #<GitHubChangelogGenerator::Generator:0x000000014edf58> (NameError)`

and after fixing that one (just changing "tag" to "tag_sha")

`/var/lib/gems/2.3.0/gems/github_changelog_generator-1.14.3/lib/github_changelog_generator/generator/generator_fetcher.rb:49:in `sha_in_tag?': Unknown status 'diverged' for comparing [my commit from march 2016]...[my commit for one tag version ago] (StandardError)`

hunner:
@zyphlar Thanks! I am testing on a repo without diverged history, so I didn't notice the misnamed variable and haven't yet tested that error condition.

For easier testing you can add my github fork to your git repo's remotes and check out the branch directly; no need to hand-copy changes. Also you can comment directly on the line of code in the files view instead of just in the comment thread making it easier to find the related line :).

zyphlar:
What do you expect to happen when there's diverged history? This is a giant
repo so maybe I can help test. This tool is only 90% useful to me while it
fails to pick up PRs when the merge is the same commit as the version...

hunner:
@zyphlar Perhaps the latest commit will fix the "merge commit is tagged" issue. And has the `tag_sha` bugfix, and uses caching to be faster now.

I still haven't handled divergent histories, which breaks how it used to work with time-based sorting, so that's coming.

olleolleolle:
(Appreciating this line of research a lot, @hunner!)

lib/github_changelog_generator/generator/generator_processor.rb
```diff
@@ -63,6 +63,36 @@ def remove_issues_in_milestones(filtered_issues)
...
+      older_tag = ensure_older_tag(older_tag, newer_tag)
+
+      issues.select do |issue|
+        # XXX It would be better to just grab merge_commit_sha but the way GCG
+        # fetches PRs currently doesn't have that info. See:
+        # https://developer.github.com/v3/pulls/#get-a-single-pull-request
+        if issue["events"] && (event = issue["events"].find { |e| e["event"] == "merged" })
+          sha = event["commit_id"]
+
+          sha_between_tags?(sha, newer_tag, older_tag)
```
hunner: This should now use the `issue["first_occurring_tag"]` as it has already been fetched.

lib/github_changelog_generator/generator/generator_fetcher.rb
```diff
@@ -15,6 +15,22 @@ def fetch_events_for_issues_and_pr
       @fetcher.fetch_events_async(@issues + @pull_requests)
     end
 
+    # Determine if a sha has occured as part of a newer tag since an older tag was made.
+    #
+    # @param sha SHA of commit to check.
+    # @param newer_tag The tag to determine if sha is a member.
+    # @param older_tag The tag prior to the newer tag.
+    # @return [Boolean] Whether the sha exists in the newer/older tag comparison.
+    def sha_between_tags?(sha, newer_tag, older_tag)
```
hunner: This is no longer needed; info is available via `add_first_occurring_tag_to_prs()`

ekohl:
When running this on [theforeman/puppet-pulp](https://github.com/theforeman/puppet-pulp)

```
github_changelog_generator -u theforeman -p puppet-pulp -o new-changelog.md --future-release 5.3.1
```

I get warnings about being unable to fetch commits:

```
Warning: Can't fetch commit ccda7a9c5899a2f73cd9736f33c0055dc7e30e5d. It is probably referenced from another repo.
Warning: Can't fetch commit f92213bec90879d940c656fe23aa4e3dfab1c829. It is probably referenced from another repo.
Warning: Can't fetch commit d1f04e13084600390a806585915fd1c9e3479f91. It is probably referenced from another repo.Warning: Can't fetch commit e86ef951db00ce4ee478155a6b670c5a95d107d7. It is probably referenced from another repo.
Warning: Can't fetch commit d6f1cb4c71e2416c639be3762ae1568786a4cd29. It is probably referenced from another repo.
Warning: Can't fetch commit 0d8f4823550555b9f2cc0bc25a8ba681ce2c384f. It is probably referenced from another repo.

Warning: Can't fetch commit d9538cf5242cffc295880120ec06cbe8db324a0d. It is probably referenced from another repo.Warning: Can't fetch commit e6debe8c630582c1c1bdf88385b1adc22a67d762. It is probably referenced from another repo.

Warning: Can't fetch commit 8d0ae9b281515f4d123496f312164ddca6da4f61. It is probably referenced from another repo.
Warning: Can't fetch commit eed8b28e2b870cee637fabb5a4449726fb51eb0a. It is probably referenced from another repo.
```

Oddly enough git show can find all commits.

hunner:
Cache invalidation allows them to be fetched, but I need to add ordering those commits by PR references. The actual "merge" commit doesn't show up for any cherry-picked commits by github's api: https://api.github.com/repos/theforeman/puppet-pulp/commits/c27e1eb vs https://api.github.com/repos/theforeman/puppet-pulp/commits/fb96de5 for https://api.github.com/repos/theforeman/puppet-pulp/pulls/282 and https://api.github.com/repos/theforeman/puppet-pulp/issues/282/events

hunner:
This should also fix #497

hunner:
And should fix #615

hunner:
And should fix #560

zyphlar:
You're doing the work of our Holiness, the Flying Spaghetti Monster
himself! Thanks ❤️

ekohl:
A quick test on a reasonably complex repository tells me this would be much better than the current code. I'd still have some issues since we've cherry picked manually so there are now github PRs to refer to. This a fundamental problem and might need metadata in a config file since there is no API that could tell us that.

hunner:
This is ready-to-go.

hunner:
**tl;dr** Me [rubberducking](http://duckie.me/) to figure out what to do about PRs that have been rebased outside of github's api.

---

The problem of rebased (traditional, or cherry-picked) merge commits not matching the API means that historically-included changes for existing changelogs would be removed from the changelog with no recourse other than to:
1. mv the current CHANGELOG.md to HISTORY.md
1. not rebase ever again

The first point isn't that bad but is annoying. The second point may be untenable to some projects. It would be nice to come up with some way to recognize rebased commits.

How could we tell?
- If there is no merge event or `merge_commit_sha` this does not always indicate the lack of a rebase merge. Eg: https://api.github.com/repos/syl20bnr/spacemacs/commits/d526121202565d1d85a8bfaf0dcbb043f38b8884 on the `develop` branch is the rebase merge of https://api.github.com/repos/syl20bnr/spacemacs/pulls/10361 . These kinds of PRs may be too ambitious to try. And these may have not ever been included anyway since they are not considered merged by github's api.
- If the `merge_commit_sha` of a given PR exists in the api but is not found in the git history, that would be the best indicator that the commit was rebased. What can we match on?
- A given rebased commit has the same `commit/author` and `status` properties... maybe.
- The commit titles may be the same, or may not.
- The commits in a merged PR may or may not have been squashed into fewer commits or had some excluded during rebase.
- The commits may reference the PR, but this would have to be done manually so there isn't any events to rely on.
- There are single-commit two-parent merges, single-commit squash merges, and single-commit non-squashed merges. Handling each of these would be hard.
- The merge commit may have two parents. Is the 2nd parent always the PR's commit?

There are too many ambiguous criteria around detecting merged commits, so what other options do we have?
- We can fallback to the time-based method like always.
- We can search the PR events for a special comment of the form "rebased commit: <sha>" that can indicate the inclusion of the PR in a tag based on a rebased sha.

The first point would mean any rebased commits would be just as wrong as they were before without a way to fix it. Not great.

The second option would solve rebased merged PRs, rebased "closed" PRs (do we want to?), cherry-picked PRs in multiple tags. Comments can be deleted or updated after merge/close to help adopting GCG for older PRs, and there can be multiple comments of this form indicating *several* tags in which the PR could be rebased for.

The second option is susceptible to outside influences of malicious users comment-spamming closed PRs with random SHAs unrelated to the PR. Threads could be locked by admins, but this is unconventional and non-default. But the submitter of a PR can affect the title already, and only admins can apply labels. Authors of PRs may spam garbage PRs, close them, and then leave comments so maybe just admins should be recognized in the events. What about if admins are removed in the future? Comments are no longer by admins. I think we can live with that.
---

Conclusion: I should add an option that checks comment events of a PR that matches `--rebase-comment` followed by a git SHA and consider those commits additional "merge" events for the PR.

ekohl:
When running with ```--unreleased-only``` it generates a ```4.0.2...5.0.1``` diff but I'd rather have the ```5.0.0...5.0.1``` diff. The repository in question is https://github.com/theforeman/puppet-candlepin which released 4.0.2 after 5.0.0 because we have some maintenance branches for released versions where we only want to cherry pick bugfixes.

hunner:
> it generates a 4.0.2...5.0.1 diff but I'd rather have the 5.0.0...5.0.1

This is because the tag comparison generation is done purely by date ordering (as most of the GCG comparisons are before this change) and the fact that the tags are on divergent branches is not currently considered.

A hash map of (branches => tags they contain) in addition to the array of time-ordered tags would be needed to resolve changelog diff comparison URLs in a realistic fashion. There isn't currently any place that needs it (though it might have helped in my shas_in_tag comparison) so I'm probably not going to add it yet.

hunner:
Rebased onto #628 for lint.

Where would docs for the rebase comment functionality go? I see https://github.com/skywinder/github-changelog-generator/wiki/Advanced-change-log-generation-examples but I'd rather have them in the codebase than on github's wiki if possible.

olleolleolle:
The man page could have a section about that, with any amount of detail. That’s a Markdown document which we use ronn to create man pages etc from.

man/git-generate-changelog.md
```diff
@@ -203,6 +203,9 @@ Automatically generate changelog from your tags, issues, labels and pull request
 
    Displays Help
 
+## REBASED COMMITS
+
+Github pull requests that have been merged whose merge commit SHA has been modified through rebasing, cherry picking, or some other method may be tracked via a special comment on Github. Git commit SHAs found in comments on pull requests matching the regular expression `/rebased commit: ([0-9a-f]{40})/i` will be used in place of the original merge SHA when being added to the changelog.
```
olleolleolle: Totally tiny thing: `Github` => `GitHub` is the on-brand spelling.

hunner:
Ready for review again. @ekohl hopefully except for the tag comparison ordering this makes your changelogs generate more correctly.

hunner:
Fixes #580

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -164,6 +164,7 @@ def fetch_closed_pull_requests
       pull_requests = []
       options = { state: "closed" }
 
+      # XXX Why does this override the base (HISTORY.md) setting?
```
hunner: This is actually the `options` hash to pass to the `@client` not the `@options` hash for gcg.

hunner:
@olleolleolle Anything you'd like me to check into for the impacts of this change? Any other feedback before merging?

lib/github_changelog_generator/generator/generator_fetcher.rb
```diff
@@ -48,6 +48,116 @@ def detect_actual_closed_dates(issues)
...
+        if pr["events"] && (event = pr["events"].find { |e| e["event"] == "merged" })
+          # Iterate tags.reverse (oldest to newest) to find first tag of each PR.
+          if (oldest_tag = tags.reverse.find { |tag| tag["shas_in_tag"].include?(event["commit_id"]) })
+            pr["first_occurring_tag"] = oldest_tag["name"]
+            found = true
+            i += 1
+            print("Associating PRs with tags: #{i}/#{total}\r") if @options[:verbose]
+          end
+        else
+          raise StandardError, "No merge sha found for PR #{pr['number']}"
```
hunner: I found a super-weird PR with regards to the github api: https://github.com/puppetlabs/puppetlabs-concat/pull/2

As mentioned there, it has no events and no merge commit. This raises an error state in the current code that I didn't expect to ever trigger. Thanks github.

hunner:
@olleolleolle Rebased again.

olleolleolle:
I’m leaning towards merging this and seeing any issues we can shake out on that side of the merge. 

The solution offered in this PR changes the unreleased version’s way of working - but it was wrong.

PR #624, opened 2018-02-21
Support printing changelog to stdout

sue445:
# Background
I want to use `github-changelog-generator` only extract PRs between latest version and current.
Then I write `CHANGELOG.md` with manual formatting. (e.g. write version up Tips, write breaking changes ...)

But `github-changelog-generator` needs file writing.
So I added a feature that print changelog to stdout without file writing.

# Example
```bash
$ github_changelog_generator --user sue445 --project my_repo --stdout --since-tag v1.2.3
(print changelog)
```

### DIFF
--- a/lib/github_changelog_generator.rb
+++ b/lib/github_changelog_generator.rb
@@ -34,10 +34,14 @@ module GitHubChangelogGenerator
     def run
       log = @generator.compound_changelog
 
-      output_filename = @options[:output].to_s
-      File.open(output_filename, "wb") { |file| file.write(log) }
-      puts "Done!"
-      puts "Generated log placed in #{Dir.pwd}/#{output_filename}"
+      if @options.write_to_file?
+        output_filename = @options[:output].to_s
+        File.open(output_filename, "wb") { |file| file.write(log) }
+        puts "Done!"
+        puts "Generated log placed in #{Dir.pwd}/#{output_filename}"
+      else
+        puts log
+      end
     end
   end
 end
--- a/lib/github_changelog_generator/options.rb
+++ b/lib/github_changelog_generator/options.rb
@@ -112,6 +112,11 @@ module GitHubChangelogGenerator
       !self[:add_sections].nil? && !self[:add_sections].empty?
     end
 
+    # @return [Boolean] whether write to `:output`
+    def write_to_file?
+      self[:output].present?
+    end
+
     private
 
     def values
--- a/lib/github_changelog_generator/parser.rb
+++ b/lib/github_changelog_generator/parser.rb
@@ -50,7 +50,7 @@ module GitHubChangelogGenerator
         opts.on("-f", "--date-format FORMAT", "Date format. Default is %Y-%m-%d") do |last|
           options[:date_format] = last
         end
-        opts.on("-o", "--output [NAME]", "Output file. Default is CHANGELOG.md") do |last|
+        opts.on("-o", "--output [NAME]", "Output file. To print to STDOUT instead, use blank as path. Default is CHANGELOG.md") do |last|
           options[:output] = last
         end
         opts.on("-b", "--base [NAME]", "Optional base file to append generated changes to.") do |last|
--- a/spec/unit/options_spec.rb
+++ b/spec/unit/options_spec.rb
@@ -40,4 +40,28 @@ RSpec.describe GitHubChangelogGenerator::Options do
       end
     end
   end
+
+  describe "#write_to_file?" do
+    subject { options.write_to_file? }
+
+    let(:options) { described_class.new(output: output) }
+
+    context "with filename" do
+      let(:output) { "CHANGELOG.md" }
+
+      it { is_expected.to eq true }
+    end
+
+    context "with nil" do
+      let(:output) { nil }
+
+      it { is_expected.to eq false }
+    end
+
+    context "with empty string" do
+      let(:output) { "" }
+
+      it { is_expected.to eq false }
+    end
+  end
 end


lib/github_changelog_generator.rb
```diff
@@ -34,10 +34,15 @@ def initialize
     def run
       log = @generator.compound_changelog
 
-      output_filename = @options[:output].to_s
-      File.open(output_filename, "wb") { |file| file.write(log) }
-      puts "Done!"
-      puts "Generated log placed in #{Dir.pwd}/#{output_filename}"
+      if @options[:stdout]
+        puts log
+        puts "Done!"
```
olleolleolle: In the case of outputting to standard out, perhaps we can allow omitting the `Done!` text?

lib/github_changelog_generator/parser.rb
```diff
@@ -185,6 +185,9 @@ def self.setup_parser(options)
         opts.on("--[no-]verbose", "Run verbosely. Default is true") do |v|
           options[:verbose] = v
         end
+        opts.on("--[no-]stdout", "Print changelog to stdout. When this is enabled, --output is ignored. Default is false") do |stdout|
+          options[:stdout] = stdout
+        end
```
olleolleolle: Another way to solve this would be to give the value `-` a special meaning in the `--output` option.

Suggestion for documentation change in `output`:

```ruby
opts.on("-o", "--output [PATH]", "Output file path. To print to STDOUT instead, use - as path. Default is CHANGELOG.md") do |last|
```

The logic in the `ChangelogGenerator` class would be about the same, but we wouldn't have to introduce and document a new option.

lib/github_changelog_generator.rb
sue445: Removed

lib/github_changelog_generator/parser.rb
sue445: I see! I added `--output -` instead of `--stdout`

sue445:
@olivierlacan I squashed my commits and `push -f`

lib/github_changelog_generator.rb
```diff
@@ -34,10 +34,14 @@ def initialize
     def run
       log = @generator.compound_changelog
 
-      output_filename = @options[:output].to_s
-      File.open(output_filename, "wb") { |file| file.write(log) }
-      puts "Done!"
-      puts "Generated log placed in #{Dir.pwd}/#{output_filename}"
+      if @options[:output]
```
olleolleolle: If the output is `""` this branch will be used. The empty string is not `false`.

Perhaps we can add a predicate method to the Options class, to be used here? Something like `#output_filename?` which would return true if a non-blank, non-`"-"` value was set.

spec/unit/options_spec.rb
```diff
@@ -40,4 +40,28 @@
...
+      let(:output) { nil }
+
+      it { is_expected.to eq false }
+    end
+
+    context "with empty string" do
+      let(:output) { "" }
+
+      it { is_expected.to eq false }
+    end
```
olleolleolle: Could we also make the case `"-"` mean write to STDOUT?

olleolleolle:
This is coming along great!

lib/github_changelog_generator.rb
sue445: Impl `#write_to_file?` at c3c45e8

> which would return true if a non-blank, non-"-" value was set.

I tried some tests, but `--output -` were not handed over to `GitHubChangelogGenerator::ChangelogGenerator#run` .

# Example
```ruby
# lib/github_changelog_generator.rb 

    def run
      log = @generator.compound_changelog

      # debug log
      p @options[:output]
```

```bash
$ bundle exec ./bin/github_changelog_generator --user sue445 --project my_repo --output - --since-tag v0.12.0
(snip)
nil

$ bundle exec ./bin/github_changelog_generator --user sue445 --project my_repo --output "-" --since-tag v0.12.0
(snip)
nil

$ bundle exec ./bin/github_changelog_generator --user sue445 --project my_repo --output "" --since-tag v0.12.0
(snip)
""

$ bundle exec ./bin/github_changelog_generator --user sue445 --project my_repo --output --since-tag v0.12.0
(snip)
nil
```

I think the hyphen is interpreted as a delimiter

spec/unit/options_spec.rb
sue445: see https://github.com/skywinder/github-changelog-generator/pull/624#discussion_r169622405

lib/github_changelog_generator/options.rb
```diff
@@ -112,6 +112,11 @@ def add_sections?
       !self[:add_sections].nil? && !self[:add_sections].empty?
     end
 
+    # Boolean method for whether write to :output
```
olleolleolle: You can express this as a YARD docblock comment: `# @return [Boolean] whether to write to :output`

lib/github_changelog_generator.rb
olleolleolle: @sue445 Oh! Such a trip-up - sorry to have led you down a wrong road.

olleolleolle:
Alright - now, only the `-` mention in the option docs for `--output` remains.

lib/github_changelog_generator/options.rb
sue445: Fix comment at 272a9a4

sue445:
@olivierlacan Removed `-` and squashed all commits

olleolleolle:
@sue445 Thank you for taking this to the finish line and your good humor in meeting all tiny change requests.

It makes this tool better suited to working with other CLI tools!

PR #626, opened 2018-02-22
Fix typo

sue445:


skywinder:
Thanks, @sue445 . Int always on our mind :) Let's *Int*ialize em all, @olleolleolle ! 👍

PR #628, opened 2018-03-05
Linty McLintface

hunner:
Looks like rubocop has a few more complaints.

### DIFF
--- a/lib/github_changelog_generator/generator/generator_tags.rb
+++ b/lib/github_changelog_generator/generator/generator_tags.rb
@@ -4,8 +4,8 @@ module GitHubChangelogGenerator
   class Generator
     # fetch, filter tags, fetch dates and sort them in time order
     def fetch_and_filter_tags
-      detect_since_tag
-      detect_due_tag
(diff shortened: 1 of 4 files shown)


lib/github_changelog_generator/generator/generator_tags.rb
```diff
@@ -99,11 +99,11 @@ def detect_link_tag_time(newer_tag)
 
     # @return [Object] try to find newest tag using #Reader and :base option if specified otherwise returns nil
     def detect_since_tag
-      @since_tag ||= options.fetch(:since_tag) { version_of_first_item }
+      @detect_since_tag ||= options.fetch(:since_tag) { version_of_first_item }
     end
 
     def detect_due_tag
-      @due_tag ||= options.fetch(:due_tag, nil)
+      @detect_due_tag ||= options.fetch(:due_tag, nil)
```
olleolleolle: Could we take the bot advice and rename the method instead? This goes for all memoized methods in this change. 

Thank you for making this change!

PR #653, opened 2018-05-11
Remove threading and use caching

### LINKED ISSUE #651: process stucks/hangs and the memory usage increases up to 4.6Gib
Hi again,
i was very happy to know that my pull request #616 got merged a few days ago so I could get rid of my fork. Sadly I figured out that using the last state of the repo the process stucks/hangs and the memory usage increased to 4.6Gib.

bildschirmfoto 2018-05-08 um 13 35 17

So I decided to use the [git-bisect](https://git-scm.com/docs/git-bisect) to figure out when was the bug introduced. As starting point i put following commit f977cd107243afa382be0f69db27119d0450985f which was added some days before I opened my fork and went the whole way through testing every step. 

    $ git bisect start
    $ git bisect bad
    $ git bisect good f977cd107243afa382be0f69db27119d0450985f

At each step i build the gem and tested agains my [project](https://github.com/sapcc/elektra/tree/master/contrib/elektra-changelog-builder) locally and not from the docker.

As a result i got following:

    372875f74838e5d7ab38bc337ec4333b540c13cd is the first bad commit
    commit 372875f74838e5d7ab38bc337ec4333b540c13cd
    Author: Hunter Haugen <hunter@puppet.com>
    Date:   Thu Feb 8 16:28:39 2018 -0800

    Use git history to find PRs in a tag instead of time

    Fixes #617

    Before this change PRs were associated with tags by the merge date of
    the PR, regardless of git branching history. In this change the oldest
    tag in which the merge commit of each PR occurs in is determined by
    using the Github compare API. That association is then used to sort PRs
    into their respective release.

    This change uses caching of commit history fetching and comparisons for
    speedup.

    :040000 040000 85ea95fe6eee626debc717ad31c2e56b790bd5bf 
    ace182352a34c7ece0d0f4c76e0e5b068d152b99 M	lib
    :040000 040000 7bf68589dad03500e971f992cfa4194824e60177 
    bd10aac7205bb84900a4a07a7ea3c3a634712e5c M	spec

I didn't had time to look into the commit. I will again open a fork rebasing to the commit before the bug and cherry pick the commits from my pull request... until we find a solution.

Best Arturo

hunner:
The octo_fetcher caches the commits, so removing the threading around methods requesting commits from the cache reduces CPU time and memory usage.

Closes #651

### DIFF
--- a/lib/github_changelog_generator/generator/generator_fetcher.rb
+++ b/lib/github_changelog_generator/generator/generator_fetcher.rb
@@ -18,34 +18,24 @@ module GitHubChangelogGenerator
     # Async fetching of all tags dates
     def fetch_tags_dates(tags)
       print "Fetching tag dates...\r" if options[:verbose]
-      # Async fetching tags:
-      threads = []
       i = 0
-      all = tags.count
       tags.each do |tag|
-        print "                                 \r"
-        threads << Thread.new do
-          get_time_of_tag(tag)
-          print "Fetching tags dates: #{i + 1}/#{all}\r" if options[:verbose]
-          i += 1
-        end
+        get_time_of_tag(tag)
+        i += 1
       end
-      threads.each(&:join)
-      puts "Fetching tags dates: #{i}" if options[:verbose]
+      puts "Fetching tags dates: #{i}/#{tags.count}" if options[:verbose]
     end
 
     # Find correct closed dates, if issues was closed by commits
     def detect_actual_closed_dates(issues)
       print "Fetching closed dates for issues...\r" if options[:verbose]
 
-      issues.each_slice(MAX_THREAD_NUMBER) do |issues_slice|
-        threads = []
-        issues_slice.each do |issue|
-          threads << Thread.new { find_closed_date_by_commit(issue) }
-        end
-        threads.each(&:join)
+      i = 0
+      issues.each do |issue|
+        find_closed_date_by_commit(issue)
+        i += 1
       end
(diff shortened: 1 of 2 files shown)


lib/github_changelog_generator/generator/generator_fetcher.rb
```diff
@@ -40,8 +40,10 @@ def detect_actual_closed_dates(issues)
...
       issues.each_slice(MAX_THREAD_NUMBER) do |issues_slice|
         threads = []
-        issues_slice.each do |issue|
-          threads << Thread.new { find_closed_date_by_commit(issue) }
+        threads << Thread.new do
+          issues_slice.each do |issue|
+            find_closed_date_by_commit(issue)
+          end
         end
         threads.each(&:join)
```
hunner: This is still wrong since the thread array and joining is inside of the `each_slice` loop. This means each thread is blocked by the join, but not the group of threads.

hunner: But when I move the threading outside of the each_slice I get memory bloat again.

hunner: Actually the original code wasn't so terrible. It just did 25 threads for the 25 issue slices, waiting for each slice. The problem seems to be threading at all because as I reduce the number of threads the memory usage *and* time to complete goes down. Removing threading makes it faster and takes less memory? Sure.

PR #686, opened 2018-10-17
Build matrix for different ruby & jruby versions

szemek:
Changes implement https://github.com/github-changelog-generator/github-changelog-generator/issues/685

To simplify configuration running Rubocop checks & RSpec tests are in shared part that is run for all Ruby & JRuby versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,26 +2,64 @@
 #
 # Check https://circleci.com/docs/2.0/language-ruby/ for more details
 #
-version: 2
+version: 2.0
+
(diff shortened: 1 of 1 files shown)


olleolleolle:
@szemek I added #687 - so, you could drop that from the matrix, reducing the size, slightly.

olleolleolle:
To get to green, #688 is perhaps needed?

szemek:
@olleolleolle I added a bit of logic that tests use different RSpec matcher depending on object class. What do you think?

PR #729, opened 2019-06-22
Make #fetch_events_async and #fetch_comments_async thread-safe

mizoR:
`iterate_page(client, ...)` is called in a thread when getting issue_events.  But `client` is not thread safe.  So sometimes changelog output can be different every time we run `github_changelog_generator`.

This PR fixs it with assigning specified octkit each thread, thanks!

### DIFF
--- a/lib/github_changelog_generator/octo_fetcher.rb
+++ b/lib/github_changelog_generator/octo_fetcher.rb
@@ -38,6 +38,7 @@ module GitHubChangelogGenerator
       prepare_cache
       configure_octokit_ssl
       @client = Octokit::Client.new(github_options)
+      @clients = (0...MAX_THREAD_NUMBER).map { Octokit::Client.new(github_options) }
     end
 
     def prepare_cache
@@ -191,10 +192,10 @@ Make sure, that you push tags to remote repo via 'git push --tags'"
       preview = { accept: Octokit::Preview::PREVIEW_TYPES[:project_card_events] }
 
       issues.each_slice(MAX_THREAD_NUMBER) do |issues_slice|
-        issues_slice.each do |issue|
-          threads << Thread.new do
+        issues_slice.each_with_index do |issue, j|
+          threads << Thread.new(@clients[j]) do |client|
             issue["events"] = []
-            iterate_pages(@client, "issue_events", issue["number"], preview) do |new_event|
+            iterate_pages(client, "issue_events", issue["number"], preview) do |new_event|
               issue["events"].concat(new_event)
             end
             issue["events"] = issue["events"].map { |event| stringify_keys_deep(event.to_hash) }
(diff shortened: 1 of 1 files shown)


alexjfisher:
@mizoR - Thank you! Thank you! Thank you, for finding and solving this!  At [Vox Pupuli](https://voxpupuli.org/), this issue has been plaguing us for months.

@ferrarimarco Can this be merged?

Thanks again!

ferrarimarco:
Great PR!

Can you please provide a test for this as well? Thanks again for the contribution :)

mizoR:
@ferrarimarco 

I tried to add test code now.

But it has not been easy for me... This is a bug that reproduces only when the response requested in the second thread arrives before the response requested in the first thread. So this is difficult to express in tests.

On the other hand, this PR is a change of `GitHubChangelogGenerator::OctoFetcher#fetch_events_async` and [`#fetch_events_async` test](https://github.com/github-changelog-generator/github-changelog-generator/blob/5630bd7/spec/unit/octo_fetcher_spec.rb#L308) passes. So could you consider merging without adding tests?

Or could you give me to any hints if you have good idea about how to test :pray:

Anyway thanks for your reply 🙂

ferrarimarco:
I think you can starting by mocking Thread.new() and work from there :)

ekohl:
@ferrarimarco I really doubt you can replicate race conditions with mocking threads.

ekohl:
Should `fetch_comments_async` use the same logic? It also uses threads. Basically grep for `MAX_THREAD_NUMBER` to find where to apply the pattern.

mizoR:
Thanks!  As far as I've grep'd, there are three functions that use threads.

- #fetch_events_async
- #fetch_comments_async
- #fetch_tag_shas_async

#fetch_events_async and #fetch_comments_async are using Octokit#last_response, so there was a race condition problem, so we fixed #fetch_comments_async as well as #fetch_events_async.

#fetch_tag_shas_async doesn't use Octokit#last_response, so I think it will be fine as is.

ioquatix:
While I think this PR looks okay... I think creates a situation where future changes of the code may introduce thread safety issues or break assumptions about the sharing of clients.

My advice is to use a thread pool and queue jobs into it:

https://github.com/ruby-concurrency/concurrent-ruby/blob/master/docs-source/thread_pools.md

In addition, unless the client is using persistent connections under the hood, caching them probably doesn't serve much purpose.

Finally, you are trying to maximise I/O throughput by putting what appears to be HTTP/1.1 requests on different threads. If that's the case, you might want to investigate async-http which would potentially improve throughput, reduce resource utilisation and avoid the need for thread synchronisation primitives.

ioquatix:
I had a quick look and octokit is using faraday:

https://github.com/octokit/octokit.rb/blob/4-stable/octokit.gemspec#L9

We have the `async-http-faraday` adapter so it should work. Just FYI.

ekohl:
@ioquatix while I agree that's a much cleaner solution, this currently fixes a nasty bug. Do you think you can come up with an alternative patch or should this be merged until a proper solution can be implemented?

ioquatix:
@ekohl https://github.com/github-changelog-generator/github-changelog-generator/pull/784 is now ready for review and should be merged in favour of this IMHO.

mizoR:
OK, I agree with that too.  I close this.  Thanks!

ioquatix:
@mizoR thanks for your work and bringing attention to this issue as well as a potential solution, we all learnt a lot!

PR #736, opened 2019-09-24
Allow generator to continue even if any errors encountered in fetch comparison

carabasdaniel:
This PR fixes issue: https://github.com/github-changelog-generator/github-changelog-generator/issues/665

There might be a more elegant solution to catch only StandardError instead of the general Exception but I think this should ensure that the change log is generated in most cases even through various errors encountered by the fetcher.

### DIFF
--- a/lib/github_changelog_generator/octo_fetcher.rb
+++ b/lib/github_changelog_generator/octo_fetcher.rb
@@ -474,6 +474,9 @@ Make sure, that you push tags to remote repo via 'git push --tags'"
       fail_with_message(e, "Exceeded retry limit")
     rescue Octokit::Unauthorized => e
       fail_with_message(e, "Error: wrong GitHub token")
+    rescue StandardError => e
+      Helper.log.error("#{e.class}: #{e.message}")
+      nil
     end
 
     # Presents the exception, and the aborts with the message.
--- a/lib/github_changelog_generator/argv_parser.rb
+++ b/lib/github_changelog_generator/argv_parser.rb
@@ -205,6 +205,9 @@ module GitHubChangelogGenerator
         opts.on("--[no-]verbose", "Run verbosely. Default is true.") do |v|
           options[:verbose] = v
         end
+        opts.on("--continue-with-errors", "Allow generator to continue even if any errors are encountered in the fetch comparison.") do |continue_with_errors|
+          options[:continue_with_errors] = continue_with_errors
+        end
         opts.on("-v", "--version", "Print version number.") do |_v|
           puts "Version: #{GitHubChangelogGenerator::VERSION}"
           exit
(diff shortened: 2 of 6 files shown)


ferrarimarco:
Hi @carabasdaniel !

Thanks for this PR.

Can you please also provide test cases?

carabasdaniel:
Hi @ferrarimarco, I've added two unit tests and amended the commit.

hzxuzhonghu:
Any update

hzxuzhonghu:
I met similar error

```
/usr/local/bundle/gems/octokit-4.14.0/lib/octokit/response/raise_error.rb:16:in `on_complete': GET https://api.github.com/repos/volcano-sh/volcano/compare/d454490cca88891c1ca8b1a85c4506d115574a21...v0.1: 404 - No common ancestor between d454490cca88891c1ca8b1a85c4506d115574a21 and v0.1. // See: https://developer.github.com/v3/repos/commits/#compare-two-commits (Octokit::NotFound)

```

genebean:
@ferrarimarco any update on this PR?

ferrarimarco:
Hi. I'm thinking about the best strategy here. While this PR let you have a changelog, it may be an issue it silently fails during a CI build where you automatically generate the changelog.

skywinder:
Thanks, @carabasdanie.
The points of @ferrarimarco are valid. 

In order to not change standard behavior we can implement this behavior with a special flag, like `--continue-with-errors`

carabasdaniel:
Hi @ferrarimarco @skywinder,

Added an option for continue_with_errors and rebased to latest master. Extracted add tag to pull request for rubocop complexity in generator fetcher.

olleolleolle:
(This code will have some movement under it, as we're waiting on #784 to land, which makes changes in the same places.)

ferrarimarco:
@carabasdaniel can you rebase against the latest master?

carabasdaniel:
Hi @ferrarimarco,

Rebased but had to remove the initially added unit tests as they were no longer valid.

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -440,6 +440,9 @@ def check_github_response
       fail_with_message(e, "Exceeded retry limit")
     rescue Octokit::Unauthorized => e
       fail_with_message(e, "Error: wrong GitHub token")
+    rescue StandardError => e
+      Helper.log.error("#{e.class}: #{e.message}")
+      nil
```
olleolleolle, line 445: 🤔 This part "skip all other errors" does not seem to be optional, and the `continue_with_errors` setting is mostly "read" at the beginning, then no more is done with it.

Is that intentional?

IrimieBogdan:
Have the same problem as @hzxuzhonghu  would love to see this PR merged!

daianamezdrea:
Same problem here...any updates about this PR?

ilopmar:
We've also hit this issue. Any updates about when this can be merged? Thanks :)

olleolleolle:
The blocker is an unmet note above:

> the `continue_with_errors` setting is mostly "read" at the beginning, then no more is done with it.

And, I can not find any test code which does things with this new setting.

Issue #738, opened 2019-10-04
TypeError: no implicit conversion of nil into Array


mcelicalderon:
I have a really weird issue on MacOS, everything works fine in Ubuntu with the same repository, I have set the same github token.

```bash
be rake changelog --trace FUTURE_RELEASE='v0.3.0'                                                                                                                                                               master
** Invoke load_app (first_time)
** Execute load_app
** Invoke changelog (first_time)
** Execute changelog
rake aborted!... 100/100
TypeError: no implicit conversion of nil into Array
/Users/mcelicalderon/.rvm/gems/ruby-2.6.5/gems/github_changelog_generator-1.14.3/lib/github_changelog_generator/octo_fetcher.rb:100:in `concat'
/Users/mcelicalderon/.rvm/gems/ruby-2.6.5/gems/github_changelog_generator-1.14.3/lib/github_changelog_generator/octo_fetcher.rb:100:in `block in github_fetch_tags'
/Users/mcelicalderon/.rvm/gems/ruby-2.6.5/gems/github_changelog_generator-1.14.3/lib/github_changelog_generator/octo_fetcher.rb:269:in `iterate_pages'
/Users/mcelicalderon/.rvm/gems/ruby-2.6.5/gems/github_changelog_generator-1.14.3/lib/github_changelog_generator/octo_fetcher.rb:97:in `github_fetch_tags'
/Users/mcelicalderon/.rvm/gems/ruby-2.6.5/gems/github_changelog_generator-1.14.3/lib/github_changelog_generator/octo_fetcher.rb:68:in `block in get_all_tags'
/Users/mcelicalderon/.rvm/gems/ruby-2.6.5/gems/github_changelog_generator-1.14.3/lib/github_changelog_generator/octo_fetcher.rb:296:in `block in check_github_response'
/Users/mcelicalderon/.rvm/gems/ruby-2.6.5/gems/retriable-2.1.0/lib/retriable.rb:54:in `block in retriable'
/Users/mcelicalderon/.rvm/gems/ruby-2.6.5/gems/retriable-2.1.0/lib/retriable.rb:48:in `times'
/Users/mcelicalderon/.rvm/gems/ruby-2.6.5/gems/retriable-2.1.0/lib/retriable.rb:48:in `retriable'
[...]

PR #739, opened 2019-10-06
Add no-http-cache option to rake task

mcelicalderon:
Simple change to allow to specify the `http_cache` option when using the rake task. This is useful as a workaround for the problem described in #738 but doesn't really fix the issue. Tested locally on my project and it worked.


ferrarimarco:
Hi @mcelicalderon! Thanks for your contribution. Would you mind adding a test case?

Thanks!

mcelicalderon:
Sure, I'll get to this as soon as possible. In fact, what would you think about having a constant with all the options available so both the task and binary can use them. That way both will always support the same options.

PR #753, opened 2020-01-21
Bugfix: undefined method line_labels_for

dusan-dragon:
If you specify option `issue-line-labels` in your `.github_changelog_generator` it will throw:
```
/usr/local/bundle/gems/github_changelog_generator-1.15.0/lib/github_changelog_generator/generator/section.rb:52:in `get_string_for_issue': undefined method `line_labels_for' for #<GitHubChangelogGenerator::Section:0x000055b07a12d628> (NoMethodError)
```

This is fix for it.

### DIFF
--- a/lib/github_changelog_generator/generator/entry.rb
+++ b/lib/github_changelog_generator/generator/entry.rb
@@ -40,6 +40,15 @@ module GitHubChangelogGenerator
       @content
     end
 
+    def line_labels_for(issue)
+      labels = if @options[:issue_line_labels] == ["ALL"]
+                 issue["labels"]
+               else
+                 issue["labels"].select { |label| @options[:issue_line_labels].include?(label["name"]) }
+               end
+      labels.map { |label| " \[[#{label['name']}](#{label['url'].sub('api.github.com/repos', 'github.com')})\]" }.join("")
+    end
+
     private
 
     # Creates section objects for this entry.
@@ -204,15 +213,6 @@ module GitHubChangelogGenerator
         @sections << merged
       end
       nil
-    end
-
-    def line_labels_for(issue)
-      labels = if @options[:issue_line_labels] == ["ALL"]
-                 issue["labels"]
-               else
-                 issue["labels"].select { |label| @options[:issue_line_labels].include?(label["name"]) }
-               end
-      labels.map { |label| " \[[#{label['name']}](#{label['url'].sub('api.github.com/repos', 'github.com')})\]" }.join("")
     end
   end
 end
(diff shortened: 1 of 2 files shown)


dusan-dragon:
@ferrarimarco ping

ferrarimarco:
Thanks for this PR! Can you please add a test case for this as well?

lib/github_changelog_generator/generator/entry.rb
```diff
@@ -40,6 +40,15 @@ def generate_entry_for_tag(pull_requests, issues, newer_tag_name, newer_tag_link
       @content
     end
 
+    def line_labels_for(issue)
```
ferrarimarco: Can you move this method where it was, so it's easier to compare with the original version?

dusan-dragon: @ferrarimarco Boss I can not. It is up here because it needs to be in `public` part. If you move it down there, this method will be private and not accessible

dusan-dragon:
ping, any update on this?

dusan-dragon:
@ferrarimarco ping

ferrarimarco:
@olleolleolle can you have a look at this as well? Thanks!

lib/github_changelog_generator/generator/section.rb
```diff
@@ -7,7 +7,7 @@ module GitHubChangelogGenerator
   #
   # @see GitHubChangelogGenerator::Entry
   class Section
-    attr_accessor :name, :prefix, :issues, :labels, :body_only
+    attr_accessor :name, :options, :prefix, :issues, :labels, :body_only
```
olleolleolle, line 10: If we're not using the setter, can `:options` become a `attr_reader`?

lib/github_changelog_generator/generator/entry.rb
olleolleolle: Right, it's called publicly later, so this can't move. Thanks, @dusan-dragon, for explaining!

lib/github_changelog_generator/generator/section.rb
```diff
@@ -49,7 +49,7 @@ def get_string_for_issue(issue)
       encapsulated_title = encapsulate_string issue["title"]
 
       title_with_number = "#{encapsulated_title} [\\##{issue['number']}](#{issue['html_url']})"
-      title_with_number = "#{title_with_number}#{line_labels_for(issue)}" if @options[:issue_line_labels].present?
+      title_with_number = "#{title_with_number}#{Entry.new(options).line_labels_for(issue)}" if @options[:issue_line_labels].present?
```
olleolleolle, line 52: Perhaps later, one can build an `@entry` to be used to repeatedly make the `line_labels_for` call on?

olleolleolle:
@ferrarimarco So, one way to deal with the code notes that I mentioned in my ✅ passing review is that we make them, when this has landed in master.

PR #770, opened 2020-03-18
Update #get_all_tags to only return tags that are for commits on the current branch's history

douglasmiller:
I've encountered a scenario where my repository has multiple branches for different versions of the library. Each major version has a branch that started off as an orphan branch.

This causes issues for me when generating the change log because the `Full Changelog` links sometimes reference a tag from the 2.x branch when I'm generating the change log for the 3.x branch.

This PR addresses that issue by modifying the `#get_all_tags` method to only include tags that are for commits that exist in the current branch's commits.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,6 +3,7 @@ PATH
   specs:
     github_changelog_generator (1.15.2)
       activesupport
+      async (>= 1.25.0)
       async-http-faraday
       faraday-http-cache
       multi_json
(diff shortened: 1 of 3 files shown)


douglasmiller:
Would it be possible to have a review of this? The issue that it resolves is a blocker for me being able to use the officially published gem and I think that it's a legitimate bug in the handling of orphan branches.

bhelx:
Would be great to get this merged. We've tested the branch and it works. This is affecting us as well. Thank you!

PR #771, opened 2020-03-18
Fixing bug when filtering pull requests without labels

douglasmiller:
The `filter_wo_labels` method is excluding pull requests that do not have labels even though the `:add_pr_wo_labels`/`pr-wo-labels` option is set to true.

This is due to an incorrect key being used to identify that the object being considered is a pull request.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -27,6 +27,10 @@ Metrics/ClassLength:
 Metrics/MethodLength:
     Enabled: false
 
+Metrics/ModuleLength:
+    Exclude:
+      - 'spec/**/*'
+
 Naming/FileName:
     Exclude:
       - 'bin/git-generate-changelog'
--- a/lib/github_changelog_generator/generator/generator_processor.rb
+++ b/lib/github_changelog_generator/generator/generator_processor.rb
@@ -159,12 +159,14 @@ module GitHubChangelogGenerator
     # @param [Array] issues Issues & PRs to filter when without labels
     # @return [Array] Issues & PRs without labels or empty array if
     #                 add_issues_wo_labels or add_pr_wo_labels are false
-    def filter_wo_labels(issues)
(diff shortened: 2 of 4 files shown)


ferrarimarco:
Hi @douglasmiller ! Can you please add a test case? Thanks!

douglasmiller:
Working on tests already

lib/github_changelog_generator/generator/generator_processor.rb
```diff
@@ -159,12 +159,14 @@ def include_issues_by_labels(issues)
     # @param [Array] issues Issues & PRs to filter when without labels
     # @return [Array] Issues & PRs without labels or empty array if
     #                 add_issues_wo_labels or add_pr_wo_labels are false
-    def filter_wo_labels(issues)
-      if (!issues.empty? && issues.first.key?("pull_requests") && options[:add_pr_wo_labels]) || options[:add_issues_wo_labels]
-        issues
-      else
-        issues.select { |issue| issue["labels"].map { |l| l["name"] }.any? }
+    def filter_wo_labels(items)
```
douglasmiller: I had to change the internals of this method beyond the initial issue of `pull_request` being incorrectly plural.

The original implementation, with the corrected value, would return all pull requests even when `:add_pr_wo_labels` was false when `:add_issues_wo_labels` was true. This was exposed when writing tests.

douglasmiller: A downside is that, when issues are passed in, the entire array will be iterated over to check for a `pull_request` key. I suppose that I could restore the `items.any? && items.first.key?("pull_request")` version if that is preferred

.rubocop.yml
```diff
@@ -27,6 +27,10 @@ Metrics/ClassLength:
 Metrics/MethodLength:
     Enabled: false
 
+Metrics/ModuleLength:
```
douglasmiller: I disabled this cop for all specs instead of the per-spec solution that was used previously. Let me know if this is not desirable.

Issue #775, opened 2020-04-02
Feature: Support Relative Tags With since_tag Option


digglife:
## Description
Under some circumstances, we need to generate changelogs for recent N versions without knowing the tag names. 

## Implementation

We can use relative tags with `since_tag` option such as `TAG~1` for since last tag, `TAG~2` for since last two tags and so on.  The syntax `TAG~1` is inspired by git (`HEAD~1`).

If the `since_tag` option value match `/TAG~([1-9]+)/`, the program will return the recent `$1` tags.

digglife:
I will add documents if it's approved by the maintainer.

PR #776, opened 2020-04-02
feat: support relative tags with since_tag option

digglife:
It's convenient if we can specify relative tags with `since_tags` option,
such as `TAG~1` for since last tag, `TAG~2` for since last 2 tags and so on.

Refer to Issue #775 for details.

### DIFF
--- a/lib/github_changelog_generator/generator/generator_tags.rb
+++ b/lib/github_changelog_generator/generator/generator_tags.rb
@@ -35,7 +35,12 @@ module GitHubChangelogGenerator
         tag = section_tags[i]
 
         # Don't create section header for the "since" tag
-        next if since_tag && tag["name"] == since_tag
+        if since_tag
+            if since_tag.match('TAG~(\d+)')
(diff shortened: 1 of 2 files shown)


digglife:
I tried to fix the code suggestions from Rubocop, but I really can't get rid of the following warning. Actually If I use  `(0..(section_tags.length - 1)).each do |i|` instead of `for i in 0..(section_tags.length - 1)`, the check would pass, so I don't think the CyclomaticComplexity check is very accurate.


```
Offenses:

lib/github_changelog_generator/generator/generator_tags.rb:32:5: C: Metrics/CyclomaticComplexity: Cyclomatic complexity for build_tag_section_mapping is too high. [8/7]
    def build_tag_section_mapping(section_tags, filtered_tags) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

30 files inspected, 1 offense detected

```

PR #786, opened 2020-04-09
Drop support for EOL Ruby versions

olleolleolle:
## Background:

Ruby 2.3, 2.4 are both EOL'd. That takes JRuby 9.1 with it, too. Read more in [Ruby Maintenance Branches](https://www.ruby-lang.org/en/downloads/branches/).

Example of 2.5 "news":
https://blog.bigbinary.com/2017/10/24/ruby-2.5-allows-rescue-inside-do-end-blocks.html

## Solution

- Raise the `required_ruby_version` to 2.5 in gemspec
- Remove CI matrix elements for unuspported Rubies.
- Remove all "version-specific" gemfiles.


ioquatix:
Yes, great idea.

olleolleolle:
In order to be able to use `async-http-faraday`, we need to be on a supported Ruby version.

Issue #756, opened 2020-01-22
Feature request: Allow to specify custom configuration file path


dusan-dragon:
So at the moment the generator is looking for config file `.github_changelog_generator` in the root of repo. It would be nice if we could override it with some cli param.

It would help us to clear a root folder a little bit and also it solve the problem if you have monorepo / multiple libs in one repository

PR #917, opened 2020-10-08
Add --config-file command line parameter

anakinj:
This addresses #756

It's still a little WIP but comments are appreciated at this point, probably going to do some refactorings.

Was thinking we could support YAML configuration files also, should be pretty straightforward to implement.

### DIFF
--- a/lib/github_changelog_generator/argv_parser.rb
+++ b/lib/github_changelog_generator/argv_parser.rb
@@ -1,0 +1,224 @@
+# frozen_string_literal: true
+
+require "optparse"
+require "github_changelog_generator/version"
+
+module GitHubChangelogGenerator
+  class ArgvParser
+    attr_reader :options
+
+    def initialize(options = {})
+      @options = options
+    end
+
+    def parse!(argv)
+      parser.parse(argv)
+    rescue OptionParser::ParseError => e
+      warn [e, parser].join("\n")
+      Kernel.abort
+    end
+
+    def parser
+      @parser ||= OptionParser.new do |opts| # rubocop:disable Metrics/BlockLength
+        opts.banner = "Usage: github_changelog_generator --user USER --project PROJECT [options]"
+        opts.on("-u", "--user USER", "Username of the owner of the target GitHub repo OR the namespace of target Github repo if owned by an organization.") do |last|
+          options[:user] = last
+        end
+        opts.on("-p", "--project PROJECT", "Name of project on GitHub.") do |last|
+          options[:project] = last
+        end
+        opts.on("-t", "--token TOKEN", "To make more than 50 requests per hour your GitHub token is required. You can generate it at: https://github.com/settings/tokens/new") do |last|
+          options[:token] = last
+        end
+        opts.on("-f", "--date-format FORMAT", "Date format. Default is %Y-%m-%d.") do |last|
+          options[:date_format] = last
+        end
+        opts.on("-o", "--output NAME", "Output file. To print to STDOUT instead, use blank as path. Default is CHANGELOG.md") do |last|
+          options[:output] = last
+        end
+        opts.on("-b", "--base NAME", "Optional base file to append generated changes to. Default is HISTORY.md") do |last|
+          options[:base] = last
+        end
+        opts.on("--summary-label LABEL", "Set up custom label for the release summary section. Default is \"\".") do |v|
+          options[:summary_prefix] = v
+        end
+        opts.on("--breaking-label LABEL", "Set up custom label for the breaking changes section. Default is \"**Breaking changes:**\".") do |v|
+          options[:breaking_prefix] = v
+        end
+        opts.on("--enhancement-label LABEL", "Set up custom label for enhancements section. Default is \"**Implemented enhancements:**\".") do |v|
+          options[:enhancement_prefix] = v
+        end
+        opts.on("--bugs-label LABEL", "Set up custom label for bug-fixes section. Default is \"**Fixed bugs:**\".") do |v|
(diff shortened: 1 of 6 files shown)


anakinj:
I think I'm done with this. Comments are appreciated.

lib/github_changelog_generator/parser.rb
```diff
@@ -6,36 +6,23 @@
 require "github_changelog_generator/helper"
 
 module GitHubChangelogGenerator
-  class Parser
-    # parse options with optparse
-    def self.parse_options
-      options = default_options
-
-      ParserFile.new(options).parse!
+  class ArgvParser
```
olleolleolle, line 9: In order to make the file layout more regular, could this class be in its own file?

olleolleolle (commented):
@anakinj Thanks for the thorough Pull Request!

I made inline comments, and you can process them as and when you wish to, let me know when you want me to look again, and I'll get back to you.

anakinj:
Thanks @olleolleolle for the super-positive-and-nice review. I will take a look at the suggestions/comments when time allows.

lib/github_changelog_generator/parser_file.rb
```diff
@@ -5,6 +5,28 @@
 module GitHubChangelogGenerator
   ParserError = Class.new(StandardError)
 
+  class FileParserChooser
+    def initialize(options, config_file)
+      @options     = options
+      @config_file = config_file
```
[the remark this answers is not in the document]
anakinj: It would. Great tip, will start using Pathname even more in the future !

lib/github_changelog_generator/parser.rb
```diff
@@ -224,51 +214,80 @@ def self.setup_parser(options)
...
+  class Parser
+    class << self
+      def parse_options(argv = ARGV)
+        options = default_options
+
+        argv_parser = ArgvParser.new(options)
+
+        argv_parser.parse!(argv) # parse first to get config_file
+        FileParserChooser.new(options, options[:config_file]).parse!
+        argv_parser.parse!(argv) # parse again to override whatever was read from the config file
```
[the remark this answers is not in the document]
anakinj: Do you mean to make the parameters even more generic to the parsers like: `FileParserChooser.new(options).parse!`

Then the parser can choose what/if it will do and the caller don't really need to care about special parameters.

lib/github_changelog_generator/parser.rb
```diff
@@ -202,6 +189,9 @@ def self.setup_parser(options)
         opts.on("--cache-file [CACHE-FILE]", "Filename to use for cache. Default is github-changelog-http-cache in a temporary directory.") do |cache_file|
           options[:cache_file] = cache_file
         end
+        opts.on("--config-file [CONFIG-FILE]", "Path to configuration file. Default is .github_changelog_generator.") do |config_file|
+          options[:config_file] = config_file
+        end
```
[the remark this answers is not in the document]
anakinj: Good catch. The surroundings impact how you do things :) Will adjust.

lib/github_changelog_generator/parser.rb
anakinj: It should be yes. I'll adjust

lib/github_changelog_generator/parser.rb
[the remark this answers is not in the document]
anakinj: Took some inspiration from this comments and changed the parsing totally from the first proposed version.

anakinj:
Great comments.

I did some adjustments and it turned out pretty nice I think.

lib/github_changelog_generator/argv_parser.rb
```diff
@@ -0,0 +1,224 @@
...
+        end
+        opts.on("-v", "--version", "Print version number.") do |_v|
+          puts "Version: #{GitHubChangelogGenerator::VERSION}"
+          exit
+        end
+        opts.on("-h", "--help", "Displays Help.") do
+          puts opts
+          exit
+        end
+      end
```
olleolleolle, line 215: When we have merged this, or earlier, we can figure out a way to not repeat (and need to maintain 2 copies of) these opts.on invocations.

lib/github_changelog_generator/parser.rb
```diff
@@ -1,277 +1,89 @@
...
   class Parser
-    # parse options with optparse
-    def self.parse_options
-      options = default_options
+    class << self
+      PARSERS = [
+        ArgvParser, # Parse arguments first to get initial options populated
+        FileParserChooser, # Then parse possible configuration files
+        ArgvParser # Lastly parse arguments again to keep the given arguments the strongest
+      ].freeze
```
olleolleolle, line 15: Nicely documented!

olleolleolle (approved):
Alright! We have gotten somewhere. Many small improvements.

Do you feel ready to merge?

lib/github_changelog_generator/argv_parser.rb
anakinj: Not totally sure I follow. But where is the copy :)?

anakinj:
Im done for now, so yeah.

lib/github_changelog_generator/argv_parser.rb
anakinj: Not totally sure I follow. But where is the copy?

olleolleolle: Ah, it was only me being momentarily confused, we're already at that place. Perfect!

olleolleolle:
@anakinj Many thanks for taking the time to build & to revisit!

PR #921, opened 2020-11-22
More concurrency

ioquatix:
Calling `#close` is no longer required, because the connection pool uses a transient task and automatically closes when going out of scope. This PR removes `#close` and also adds a top level scope around the generator, so that persistent connections can be reused between different parts of `OctoFetcher`.


ioquatix:
If someone has time to compare the perf, it could be interesting. I'm not sure it's a big win, it's more about the simplicity of the code.

olleolleolle (approved):
Thank you, this small change – with the Sync block – is way easier to follow ("Where does it end? Ah, at `end`!").

PR #926, opened 2020-12-28
Use UTC for future release date

smortex:
When using future releases, the current local date is writen in the
generated CHANGELOG.  This is usualy done as part of releasing a new
version, that is a few seconds before a that version is tagged and
that tag is pushed to GitHub.

On next version bump, that previous tag exist and it's date is read from
GitHub which provide this information in Coordinated Universal Time
(UTC).

Depending on the time-zone of the user, the date of these two Ruby Time
objects may be different, leading to an unexpected change in the
previous entry of the CHANGELOG.

To avoid this, convert the local time to UTC when we do not have an
actual tag time to use.

### DIFF
--- a/spec/unit/generator/generator_tags_spec.rb
+++ b/spec/unit/generator/generator_tags_spec.rb
@@ -13,6 +13,33 @@ describe GitHubChangelogGenerator::Generator do
     end
   end
 
+  describe "#detect_link_tag_time" do
+    let(:newer_tag) { nil }
+
+    let(:default_options) { GitHubChangelogGenerator::Parser.default_options.merge(verbose: false) }
+    let(:options) do
+      {
+        future_release: "2.0.0"
+      }
+    end
+    let(:generator) { described_class.new(default_options.merge(options)) }
+
+    subject do
+      generator.detect_link_tag_time(newer_tag)
+    end
+
+    context "When the local date is not the same as the UTC date" do
+      before do
+        # 2020-12-27T17:00:00-10:00 is 2020-12-28T03:00:00Z.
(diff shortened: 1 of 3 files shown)


spec/unit/generator/generator_tags_spec.rb
```diff
@@ -13,6 +13,28 @@ def tags_from_strings(tags_strings)
...
+    subject do
+      generator.detect_link_tag_time(newer_tag)
+    end
+
+    before do
+      expect(Time).to receive(:new).and_return(Time.new(2020, 12, 27, 17, 0, 0, "-10:00"))
+    end
+
+    it { is_expected.to eq(["2.0.0", "2.0.0", Time.gm(2020, 12, 28, 3)]) }
+  end
```
olleolleolle, line 36: Thanks for adding a spec, this helps catch regressions. But, perhaps a code comment, or an "it" being described with a note that captures the essence of the change (should always use UTC since GitHub uses UTC for the things we care about, etc).

olleolleolle (commented):
Thanks a lot, this looks like a quality of life change!

spec/unit/generator/generator_tags_spec.rb
smortex: lol, I was focusing what was being tested and the reason for this has been completely lost!

I amended my commit to add a context and an explicit comment.  Thanks!

olleolleolle: This looks perfect, and will guide the interested reader in the future.

Thanks, @smortex for the repairs!

olleolleolle (approved):
LGTM!

Issue #376, opened 2016-05-11
[wishlist] --include-tags-regex


yarikoptic:
There is `--exclude-tags-regex` but there could be a variety of non-release tags in the repository.  So IMHO it would be a more common use-case to be able to specify which tags regex to consider (instead of trying to catch all the ones to exclude).  Or is there a "negation" which could be specified within exclude-tags-regex, so I could say "exclude all which do not match this regex".  Thanks in advance

donquixote:
Yes please!
Typical use case:
- include semver tags with include-tags-regex
- exclude specific semver tags with exclude-tags-regex

donquixote:
Fixed by #929 ?

PR #929, opened 2021-01-15
Add --include-tags-regex option.

ameir:
For my usecase, excluding tags would be too cumbersome. This PR allows for the ability to do things like `--include-tags-regex='v0.[56]'` to focus on a certain range, or `--include-tags-regex='^v\d+.\d+.\d+$'` to exclude all tags with `-develop` or `-rc` or similar suffixes.


ccannell67 (approved):
LGTM

olleolleolle (approved):
Solid addition, with test and the implementation looks straight-forward.

First, filter what to include, then filter out what to exclude from that. So, it is possible to use both together, very clear.

donquixote:
Seems that this fixed #376

PR #946, opened 2021-03-26
Update help output to reflect required args.

### LINKED ISSUE #945: Not all options listed in OptionParser have optional arguments
**Describe the bug**

The listed options are routinely shown as having `--optname [FLORP]` where FLORP is now described as _optional_ argument to the `--optname`.

This is a docs problem, since many of these require a value to work correctly.

**To Reproduce**
1. Go to lib/github_changelog_generator/parser.rb
2. Read the individual options defined
3. See issue that many are `[OPTIONAL-VALUE]` when they should not.

**Expected behavior**

`--help` output which matches what is true for option arity.

**Additional context**

This is a ✨ 📖 DOCUMENTATION ISSUE 📖 ✨ .

spark-c:
This PR fixes #945

Help output messages should now properly reflect whether args are required.
I didn't find any Options which seemed to take optional arguments; they all look like they require args OR require zero args.

Another pair of eyes would be great to double-check!

### DIFF
--- a/man/git-generate-changelog.1
+++ b/man/git-generate-changelog.1
@@ -13,115 +13,115 @@
 Automatically generate changelog from your tags, issues, labels and pull requests on GitHub\.
 .
 .SH "OPTIONS"
-\-u, \-\-user [USER]
+\-u, \-\-user USER
 .
 .P
 Username of the owner of target GitHub repo\.
 .
 .P
-\-p, \-\-project [PROJECT]
+\-p, \-\-project PROJECT
 .
 .P
 Name of project on GitHub\.
 .
 .P
-\-t, \-\-token [TOKEN]
+\-t, \-\-token TOKEN
 .
 .P
 To make more than 50 requests per hour your GitHub token is required\. You can generate it at: https://github\.com/settings/tokens/new
 .
 .P
-\-f, \-\-date\-format [FORMAT]
+\-f, \-\-date\-format FORMAT
 .
 .P
 Date format\. Default is %Y\-%m\-%d\.
 .
 .P
-\-o, \-\-output [NAME]
+\-o, \-\-output NAME
 .
 .P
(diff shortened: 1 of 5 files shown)


olleolleolle (commented):
Cool, the additional notation looks right. 

There are a few documentation files which repeat this info, if you can find them, perhaps in man/ or doc/ or someplace, hunt for it - if you aren't able to get them updated, just say so, and we'll merge this change as a step on the way.

This will change the --help output for the user trying the tool interactively. The man page is for a smaller, discerning audience, who read that stuff. (Same content, hehe.)

Let me know if you are up to looking for how to regenerate or edit those files!

spark-c:
Sure thing! I'll do some digging around and get those updated as well.

spark-c:
Okay! Not sure how I missed the man/ directory the first time around haha.

The four man files in that directory have now been updated.

man/git-generate-changelog.1
```diff
@@ -13,115 +13,115 @@
...
 .P
-\-\-configure\-sections [HASH, STRING]
+\-\-configure\-sections HASH, STRING
 .
 .P
 Define your own set of sections which overrides all default sections\.
 .
 .P
-\-\-add\-sections [HASH, STRING]
+\-\-add\-sections HASH, STRING
```
olleolleolle, line 112: Maybe we should investigate this. Me, reading it, can't from the example figure out real quick what I can write here. Is Hash defined anywhere? How do I tell the CLI to enter a hash here? Can I do that in the CLI version?

I understand that this was not introduced by your amazing changes, but perhaps we need to capture this in an Issue.

olleolleolle (approved):
Thanks for fixing this!

It improves the quality of the documentation a lot!

olleolleolle:
@spark-c Thank you!

Issue #947, opened 2021-03-26
Allow `bin/github_changelog_generator` to be invoked from a path outside the repo


magneland:
**Is your feature request related to a problem? Please describe.**
As part of my local development workflow, I would like to make some code changes to this repo, and quickly run the generator on another gem for testing it out.
I tried this command locally (note the `cwd` is outside the forked repo):

```
magne@C02ZV1LNMD6W af_json_api % /Users/magne/github/magneland/github-changelog-generator/bin/github_changelog_generator --base CHANGELOG.md --token TOKEN --user USER --project af_json_api
Traceback (most recent call last):
	4: from /Users/magne/github/magneland/github-changelog-generator/bin/github_changelog_generator:4:in `<main>'
	3: from /Users/magne/github/magneland/github-changelog-generator/bin/github_changelog_generator:4:in `require_relative'
	2: from /Users/magne/github/magneland/github-changelog-generator/lib/github_changelog_generator.rb:13:in `<top (required)>'
	1: from /Users/magne/.rvm/rubies/ruby-2.7.1/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:155:in `require'
/Users/magne/.rvm/rubies/ruby-2.7.1/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:155:in `require': cannot load such file -- github_changelog_generator/helper (LoadError)
```

The error happens because files in `lib/` use the "require" directive to include other files in `lib/`.

Incidentally, `bin/github_changelog_generator` already makes use of `require_relative` so that is already a pattern.

**Describe the solution you'd like**
Use `require_relative` instead of `require`.

**Describe alternatives you've considered**
Not sure if there's a better way or a workaround?

**Additional context**
See: https://github.com/rubocop/rubocop/issues/8748

olleolleolle:
Have you attempted to use the Bundler feature "local"? It allows Bundler to recognize a path on your computer as "the" resolution for a gem in a Gemfile. 


If you add a Gemfile to the gem you wish to test (it probably exists already), and then, add a gem "github_changelog_generator" in there, when you run the command in the "other gem" project, Bundler will pick your cool-and-edited work instead of "whatever files it otherwise would find". 

Another way is to use "bundle open" on this gem, and edit its files in place, and then undo those changes when done.

[...]

PR #948, opened 2021-03-26
Replace "require" with "require_relative" where possible

magneland:
Allows `bin/github_changelog_generator` to be invoked from anywhere.
This facilitates easier ad hoc testing locally.

This commit leaves specs unchanged on purpose.

See: https://github.com/rubocop/rubocop/issues/8748


magneland:
Fixes #947

Issue #928, opened 2021-01-11
Add Ruby 3 to CI, make it pass


olleolleolle:
- add Ruby 3 to https://github.com/github-changelog-generator/github-changelog-generator/blob/master/.circleci/config.yml
- perhaps remove the version specifier in the `gem install bundler` section, latest is fine
- use newer ruby Orb `circleci/ruby@1.0`
- make it pass tests without warnings or failures
- or, if that's what you want, pick GitHub Actions instead?

PR #949, opened 2021-03-26
Support Ruby 3

magneland:


magneland:
Ruby 3 contains a number of changes:
https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/

However, the only relevant change appears to be:
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

This PR changes the CircleCI config to support Ruby 3, but also refactors the config.
It's unclear to me why the old checks are still Required - maybe due to some GitHub project settings?

lib/github_changelog_generator/generator/generator_processor.rb
```diff
@@ -156,7 +156,7 @@ def include_issues_by_labels(issues)
       filter_wo_labels(filtered_issues)
     end
 
-    # @param [Array] issues Issues & PRs to filter when without labels
+    # @param [Array] items Issues & PRs to filter when without labels
```
magneland, line 159: It's debatable whether this commit belongs in this PR.
I would probably prefer a Rubocop cop or something that auto-corrects this.

magneland (commented):
Fixes #928

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -382,6 +388,7 @@ def commits_in_tag(sha, shas = Set.new)
       shas
     end
 
+    # @param [Object] indata
```
olleolleolle, line 391: These are super cool additions, and type descriptions in Yard can use a comma to separate each matching type like [Array, Hash] etc.

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -442,6 +452,7 @@ def iterate_pages(client, method, *arguments, parent: nil, **options)
     # This is wrapper with rescue block
     #
     # @return [Object] returns exactly the same, what you put in the block, but wrap it with begin-rescue block
+    # @param [Proc] block
```
olleolleolle, line 455: Yard probably has some notation for you here, specific to blocks. We don't want the docs to show that this method takes a positional argument. Does the rendered Yard docs HTML show that?

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -483,6 +496,7 @@ def retry_callback
       end
     end
 
+    # @param [Object] msg
```
olleolleolle, line 499: This type is String.

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -453,6 +464,8 @@ def check_github_response(&block)
     end
 
     # Presents the exception, and the aborts with the message.
+    # @param [Object] message
+    # @param [Object] error
```
olleolleolle, line 468: Error is Exception I guess.

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -453,6 +464,8 @@ def check_github_response(&block)
     end
 
     # Presents the exception, and the aborts with the message.
+    # @param [Object] message
```
olleolleolle, line 467: Message is a String.

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -405,10 +412,13 @@ def stringify_keys_deep(indata)
     #
     # @param [Octokit::Client] client
     # @param [String] method (eg. 'tags')
+    # @param [Array] arguments
+    # @param [Async::Semaphore] parent
     #
     # @yield [Sawyer::Resource] An OctoKit-provided response (which can be empty)
     #
     # @return [void]
+    # @param [Hash] options
```
olleolleolle, line 421: Minor: this param should be placed before th return Yard directive.

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -357,6 +361,8 @@ def fetch_tag_shas(tags)
 
     private
 
+    # @param [Set] shas
+    # @param [Object] sha
```
olleolleolle, line 365: The sha is String, I think.

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -335,6 +338,7 @@ def default_branch
       @default_branch ||= client.repository(user_project)[:default_branch]
     end
 
+    # @param [Object] name
```
olleolleolle, line 341: Name is String.

lib/github_changelog_generator/generator/generator_processor.rb
```diff
@@ -170,6 +170,7 @@ def filter_wo_labels(items)
     end
 
     # @todo Document this
+    # @param [Object] issues
```
olleolleolle, line 173: This enumerable is probably an Array

lib/github_changelog_generator/generator/generator_processor.rb
olleolleolle: Yard-junk is a tool for this. Finding errors and presenting them in a small report.

olleolleolle:
Hi, I made copious inline comments. 

Also: yay! A great move to begin using the Ruby Orb better.

olleolleolle:
In order to use this, I will move forward without fixing the pointed-out YARD annotations.

magneland:
> In order to use this, I will move forward without fixing the pointed-out YARD annotations.

Cool thanks, makes sense. I saw your comments about YARD but did not have time to go through them all.

PR #966, opened 2021-06-01
Feat: Allow named markdown anchor links

kikito:
Markdown links don't always have their urls attached. Named links allow
putting them afterwards. This change allows the reader to understand
this pattern on the tag headers:

    ## [1.0.0]

    This is version 1.0.0

    ## [0.9.1]

    This is version 0.9.1

    ## [0.9.0]
    ...

    [1.0.0]: https://github.com/myorg/myrepo/compare/0.9.1...1.0.0
    [0.9.1]: https://github.com/myorg/myrepo/compare/0.9.1...0.9.0
    ...

Without this change, the generator will get confused by the brackets
around the versions, and complain that "it cannot find the [1.0.0]
tag" on `all_tags` (which don't have the brackets)

### DIFF
--- a/lib/github_changelog_generator/reader.rb
+++ b/lib/github_changelog_generator/reader.rb
@@ -29,7 +29,7 @@ module GitHubChangelogGenerator
       defaults = {
         heading_level: "##",
         heading_structures: [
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
The tests empower the reviewer, as tbray said.

This change would help the user who wanted to link to a version, as well.

PR #967, opened 2021-06-02
Directly handle rate limiting using `client.rate_limit.resets_in`.

ioquatix:
We discussed some issues here, and at least when I tried, it, I ran into rate limiting issues.

https://github.com/github-changelog-generator/github-changelog-generator/discussions/965#discussioncomment-816764

This PR tries to respect GitHub's rate limiting.

### DIFF
--- a/lib/github_changelog_generator/octo_fetcher.rb
+++ b/lib/github_changelog_generator/octo_fetcher.rb
@@ -463,7 +463,7 @@ Make sure, that you push tags to remote repo via 'git push --tags'"
       resets_in = client.rate_limit.resets_in
       Helper.log.error("#{e.class} #{e.message}; sleeping for #{resets_in}s...")
 
-      if task = Async::Task.current?
+      if (task = Async::Task.current?)
         task.sleep(resets_in)
       else
         sleep(resets_in)
--- a/lib/github_changelog_generator/octo_fetcher.rb
+++ b/lib/github_changelog_generator/octo_fetcher.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require "tmpdir"
-require "retriable"
 require "set"
 require "async"
 require "async/barrier"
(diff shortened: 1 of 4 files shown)


olleolleolle:
Ugh I was so curious about the test results, and it was styling issues that stopped the build. I should flip those around, check style after running tests.

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -456,10 +455,21 @@ def iterate_pages(client, method, *arguments, parent: nil, **options)
...
     # @return [Object] returns exactly the same, what you put in the block, but wrap it with begin-rescue block
     # @param [Proc] block
-    def check_github_response(&block)
-      Retriable.retriable(retry_options, &block)
+    def check_github_response
+      yield
     rescue MovedPermanentlyError => e
       fail_with_message(e, "The repository has moved, update your configuration")
+    rescue Octokit::TooManyRequests => e
+      resets_in = client.rate_limit.resets_in
```
olleolleolle, line 463: 🔍 I learned! resets_in (a time) is even provided by Octokit, thanks for finding this out!

olleolleolle (approved):
👍 This lets us use Octokit in a more precise way.

lib/github_changelog_generator/octo_fetcher.rb
ioquatix: You are most welcome!

ioquatix:
Thanks for handling this!

aelam:
it seems that It does not work well as before.

ioquatix:
@aelam what is the problem?

aelam:
@ioquatix Sorry for late, after I waited a long time, it worked

Mizommz:
> We discussed some issues here, and at least when I tried, it, I ran into rate limiting issues.
> 
> https://github.com/github-changelog-generator/github-changelog-generator/discussions/965#discussioncomment-816764
> 
> This PR tries to respect GitHub's rate limiting.

lib/github_changelog_generator/octo_fetcher.rb
Mizommz: https://androidonair.withgoogle.com/events/android-security?talk=session-2

PR #984, opened 2021-10-02
Extract GitHubChangelogGenerator::FileParserChooser to its own file

anakinj:
Minor change that moves the FileParserChooser class into it's own file.

### DIFF
--- a/lib/github_changelog_generator/file_parser_chooser.rb
+++ b/lib/github_changelog_generator/file_parser_chooser.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require "pathname"
+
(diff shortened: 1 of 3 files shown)


lib/github_changelog_generator/file_parser_chooser.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require "pathname"
+
+module GitHubChangelogGenerator
+  class FileParserChooser
+    def initialize(options)
+      @options     = options
```
olleolleolle, line 8:
```suggestion
      @options = options
```

olleolleolle (approved):
Thanks!

PR #985, opened 2021-10-02
refactor: Unnest conditionals in tag-related methods, deprecate an ill-named method

anakinj:
Took a little time to refactor a few tag related methods, inspired by removing the Cop `Style/GuardClause` from the RuboCop ignore list.

### DIFF
--- a/lib/github_changelog_generator/generator/generator_fetcher.rb
+++ b/lib/github_changelog_generator/generator/generator_fetcher.rb
@@ -160,7 +160,7 @@ module GitHubChangelogGenerator
       compare_string = issue["merged_at"].nil? ? "closed" : "merged"
       # reverse! - to find latest closed event. (event goes in date order)
       issue["events"].reverse!.each do |event|
-        if event["event"].eql? compare_string
+        if event["event"] == compare_string
           set_date_from_event(event, issue)
           break
         end
--- a/lib/github_changelog_generator/generator/generator_fetcher.rb
+++ b/lib/github_changelog_generator/generator/generator_fetcher.rb
@@ -154,15 +154,15 @@ module GitHubChangelogGenerator
     # Fill :actual_date parameter of specified issue by closed date of the commit, if it was closed by commit.
     # @param [Hash] issue
     def find_closed_date_by_commit(issue)
-      unless issue["events"].nil?
-        # if it's PR -> then find "merged event", in case of usual issue -> fond closed date
-        compare_string = issue["merged_at"].nil? ? "closed" : "merged"
-        # reverse! - to find latest closed event. (event goes in date order)
-        issue["events"].reverse!.each do |event|
-          if event["event"].eql? compare_string
-            set_date_from_event(event, issue)
-            break
-          end
+      return if issue["events"].nil?
+
+      # if it's PR -> then find "merged event", in case of usual issue -> fond closed date
+      compare_string = issue["merged_at"].nil? ? "closed" : "merged"
+      # reverse! - to find latest closed event. (event goes in date order)
+      issue["events"].reverse!.each do |event|
+        if event["event"].eql? compare_string
+          set_date_from_event(event, issue)
+          break
         end
       end
       # TODO: assert issues, that remain without 'actual_date' hash for some reason.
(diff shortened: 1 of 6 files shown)


lib/github_changelog_generator/generator/generator_fetcher.rb
```diff
@@ -154,15 +154,15 @@ def associate_rebase_comment_prs(tags, prs_left, total)
...
-        compare_string = issue["merged_at"].nil? ? "closed" : "merged"
-        # reverse! - to find latest closed event. (event goes in date order)
-        issue["events"].reverse!.each do |event|
-          if event["event"].eql? compare_string
-            set_date_from_event(event, issue)
-            break
-          end
+      return if issue["events"].nil?
+
+      # if it's PR -> then find "merged event", in case of usual issue -> fond closed date
```
olleolleolle, line 159: Unrelated spelling fix in comment

```suggestion
      # if it's PR -> then find "merged event", in case of usual issue -> found closed date
```

lib/github_changelog_generator/generator/generator_fetcher.rb
```diff
@@ -154,15 +154,15 @@ def associate_rebase_comment_prs(tags, prs_left, total)
...
+      compare_string = issue["merged_at"].nil? ? "closed" : "merged"
+      # reverse! - to find latest closed event. (event goes in date order)
+      issue["events"].reverse!.each do |event|
+        if event["event"].eql? compare_string
```
olleolleolle, line 163: Unrelated to this refactoring: here we can use == instead of eql?, I believe:

Blog post, on the difference:
https://medium.com/@khalidh64/difference-between-eql-equal-in-ruby-2ffa7f073532

lib/github_changelog_generator/generator/generator_tags.rb
```diff
@@ -207,14 +195,16 @@ def filter_exact_tags(all_tags)
     end
 
     def warn_if_nonmatching_regex(all_tags, regex, regex_option_name)
-      unless all_tags.map { |t| t["name"] }.any? { |t| regex =~ t }
-        Helper.log.warn "Warning: unable to reject any tag, using regex "\
-                        "#{regex.inspect} in #{regex_option_name} option."
-      end
+      return if all_tags.any? { |t| regex =~ t["name"] }
```
olleolleolle, line 198: Can we use `match?` here?

lib/github_changelog_generator/generator/generator_tags.rb
```diff
@@ -119,41 +119,29 @@ def get_filtered_tags(all_tags)
...
-          idx = all_tags.index { |t| t["name"] == tag }
-          filtered_tags = if idx > 0
-                            all_tags[(idx + 1)..-1]
-                          else
-                            []
-                          end
-        else
-          raise ChangelogGeneratorError, "Error: can't find tag #{tag}, specified with --due-tag option."
-        end
+      return all_tags unless (tag = due_tag)
```
olleolleolle, line 136: This here is a very compact fix to the snarl of conditionals in the before. Kudos.

olleolleolle (approved):
**Thanks** for renovating! 🎉

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -95,7 +95,7 @@ def client
     # Fetch all tags from repo
     #
     # @return [Array <Hash>] array of tags
-    def get_all_tags
+    def fetch_all_tags
```
olleolleolle, line 98: Since this is a public API, even re-imagined by a neighbouring effort to use this codebase on GitLab repos, can we retain an alias to get_all_tags, to make it compatible? 

I like this rename, _and_ I'd like to make a compatibility alias.

olleolleolle: If you like, you can emit a deprecation warning in the get_all_tags method.

anakinj: Good catch. Brought back the old method with a deprecation warning.

lib/github_changelog_generator/generator/generator_tags.rb
anakinj: yes

lib/github_changelog_generator/generator/generator_fetcher.rb
anakinj: Think `==` is what we want.

PR #986, opened 2021-10-02
Fix for echoing the complete output path

### LINKED ISSUE #892: Generated log path incorrect
Hi, this is a great project! It took me a while to get it up and running but now I've got it going I can see how useful it will be in the future for my company's projects. :) 

I'm using it within Docker, and have got a changelog generating using the following command: 

```
docker run -it --rm \
  -v ~/projects/project-name:/data \
  ferrarimarco/github-changelog-generator \
  github_changelog_generator \
  --user MyOrganization \
  --project Project-Name \
  --token ***
```

When the changelog gets generated, though, the script prints that the log has been placed in the wrong directory: 

```
Generating entry...
Done!
Generated log placed in /usr/local/src/your-app/CHANGELOG.md
```

When I change the `--output` value eg to `/data/CHANGELOG.md` then the path gets updated to `/usr/local/src/your-app//data/CHANGELOG.md`. I'm guessing that this is because the script is running within Docker and it prints `$(pwd)` followed by the output path.

The changelog does get generated in the right place, but the path is incorrect.

anakinj:
This should fix #892 and also extends test coverage to the entrypoint of the app to verify the behaviour.


spec/github_changelog_generator_spec.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+describe GitHubChangelogGenerator::ChangelogGenerator do
```
olleolleolle, line 3: In this project, prefer the non-monkey-patched form:

```suggestion
RSpec.describe GitHubChangelogGenerator::ChangelogGenerator do
```

Perhaps we can enable config.disable_monkey_patching! for RSpec.

spec/tmp/.gitignore
```diff
@@ -0,0 +1 @@
+*
```
olleolleolle, line 1: Can this be a part of the .gitignore in the root?

lib/github_changelog_generator.rb
```diff
@@ -24,24 +24,28 @@ module GitHubChangelogGenerator
   class ChangelogGenerator
     # Class, responsible for whole changelog generation cycle
     # @return initialised instance of ChangelogGenerator
-    def initialize
-      @options = Parser.parse_options
-      @generator = Generator.new @options
+    def initialize(params = ARGV)
+      @options   = Parser.parse_options(params)
```
olleolleolle, line 28:
```suggestion
      @options = Parser.parse_options(params)
```

lib/github_changelog_generator.rb
```diff
@@ -24,24 +24,28 @@ module GitHubChangelogGenerator
...
-        puts "Generated log placed in #{Dir.pwd}/#{output_filename}"
+        puts "Generated log placed in #{File.absolute_path(output_filename)}"
       else
         puts log
       end
     end
+
+    private
+
+    attr_reader :generator
```
olleolleolle, line 49: If we are going to have private readers, options wants one, too.

spec/tmp/.gitignore
anakinj: The file serves two purposed. It ensures the repository has a `tmp` folder for the specs and it ensures there is never any files in the folder. I could be in the root .gitignore, but then there would be some kind of `.keep` to ensure the foder would be in git.

lib/github_changelog_generator.rb
anakinj: Either it was private reader of playing around with `instance_variable_set` in the tests. So went with this one to be able to mock the generator. I'll go ahead and use private readers for all the instance variables.

olleolleolle (approved):
Thanks, @anakinj!

PR #999, opened 2021-11-09
document how to create release notes

bruce-szalwinski:
This adds a new section to the README, "Practical Use Cases" that documents how to combine `github_changelog_generator` with github cli to create release notes.   I used `doctoc` to update the table of contents.  Looks like a couple of sections moved since the last time the table of contents was updated.


olleolleolle (approved):
Thanks! This is a carefully added note that makes room for other such notes. 👍 🎉

PR #1004, opened 2022-01-05
Add contributors on own line option

boardfish:
Hey, folks! I'm investigating adopting this for ViewComponent. As I identified [here](https://github.com/github/view_component/issues/1182#issuecomment-1005002162), we'd need a few tweaks to do so and reflect the current changelog. This is one of them - my implementation feels a little bit sketchy so far, so I'd appreciate any advice I can get on how to comfortably have this sit alongside the current code.

### DIFF
--- a/lib/github_changelog_generator/argv_parser.rb
+++ b/lib/github_changelog_generator/argv_parser.rb
@@ -205,6 +205,9 @@ module GitHubChangelogGenerator
         opts.on("--[no-]verbose", "Run verbosely. Default is true.") do |v|
           options[:verbose] = v
         end
+        opts.on("--contributors-on-own-line", "Place contributors on their own line in italics. Default is false.") do |v|
+          options[:verbose] = v
+        end
         opts.on("-v", "--version", "Print version number.") do |_v|
           puts "Version: #{GitHubChangelogGenerator::VERSION}"
           exit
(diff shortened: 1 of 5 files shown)


lib/github_changelog_generator/generator/section.rb
```diff
@@ -97,14 +98,18 @@ def body_till_first_break(body)
     def issue_line_with_user(line, issue)
       return line if !@options[:author] || issue["pull_request"].nil?
 
-      user = issue["user"]
+      user = user(issue["user"])
       return "#{line} ({Null user})" unless user
```
yykamei, line 102: `GitHubChangelogGenerator::Section#user` seems to only return `String` as far as I see, so I think this line is no longer needed.

```suggestion
```

yykamei (commented):
Hi, I'm not a maintainer, but I'm interested in this pull request. In my opinion, It looks good 👍 but I left a trivial comment.

I wonder what's next if it's ready to merge 👀

lib/github_changelog_generator/argv_parser.rb
```diff
@@ -205,6 +205,9 @@ def parser
         opts.on("--[no-]verbose", "Run verbosely. Default is true.") do |v|
           options[:verbose] = v
         end
+        opts.on("--contributors-on-own-line", "Place contributors on their own line in italics. Default is false.") do |v|
+          options[:verbose] = v
```
skywinder: This assigns the new flag to `:verbose` instead of `:contributors_on_own_line`, so `--contributors-on-own-line` does not enable the feature and unexpectedly toggles verbose output. The parser needs to write the new option key here.

PR #1032, opened 2023-10-01
Update to Ruby 3.2.2, raise lowest-supported Ruby version to 3.0

mencarellic:
[Ruby 2.7 went EOL in March of 2023](https://www.ruby-lang.org/en/downloads/branches/). This brings the codebase up to Ruby 3.2.2 with the relevant changes.

I tested this against the [changelog_test](https://github.com/github-changelog-generator/changelog_test) repo and saw what I believe were the correct results.


olleolleolle:
My preference would be "drop the ones that are fully out of support", and bring new ones in.

github_changelog_generator.gemspec
```diff
@@ -8,7 +8,7 @@ Gem::Specification.new do |spec|
   spec.name               = "github_changelog_generator"
   spec.version            = GitHubChangelogGenerator::VERSION
 
-  spec.required_ruby_version = ">= 2.5.0"
+  spec.required_ruby_version = ">= 3.2.0"
```
olleolleolle, line 11: Oh, can we keep this to 3.0.0? This is the required lowest version that's allowed to run the software.

mencarellic: Sure. I'll make some adjustments since it'll affect the cops enabled.

lib/github_changelog_generator/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module GitHubChangelogGenerator
-  VERSION = "1.16.4"
+  VERSION = "1.17.0"
```
olleolleolle, line 4:
```suggestion
  VERSION = "1.16.4"
```

The release process takes care of updating the version number. Don't change this in non-release PRs.

olleolleolle: No biggie, I can resolve this.

PR #1039, opened 2024-02-06
Handle Null Comments Result

devinlyons:
This was causing issues with one of my repos. This seems to have fixed it.

### DIFF
--- a/lib/github_changelog_generator/octo_fetcher.rb
+++ b/lib/github_changelog_generator/octo_fetcher.rb
@@ -254,7 +254,7 @@ Make sure, that you push tags to remote repo via 'git push --tags'"
           semaphore.async do
             pr["comments"] = []
             iterate_pages(client, "issue_comments", pr["number"]) do |new_comment|
-              pr["comments"].concat(new_comment)
+              pr["comments"].concat(new_comment) if new_comment
             end
             pr["comments"] = pr["comments"].map { |comment| stringify_keys_deep(comment.to_hash) }
           end


devinlyons:
If this gets merged, what would it take to get a new release? I would like to update https://github.com/github-changelog-generator/docker-github-changelog-generator too.

olleolleolle:
@devinlyons I can see that this would fix that issue, can you figure out a test case that could go in the repo, too?

devinlyons:
@olleolleolle, my Ruby skills are a bit weak. But I did look through the tests to see if I could find an example to work from. It looks like the `fetch_comments_async` function is only ever mocked in the tests. I guess this makes sense given that it does a lot of IO. So, I'm not sure how to test this. The use case is something like `it does not throw an exception when a page returns no comments`. Do you have any guidance on this?

olleolleolle:
I took a look at the yield which this conditional is protecting from.

Was curious if the `#iterate_pages` could be made so that it didn't yield if it got a nil `last_response.data` (which is used in `yield(last_response.data)`) in `#iterate_pages`.

`# @yield [Sawyer::Resource] An OctoKit-provided response (which can be empty)`

I also couldn't see a quick way to test-drive that thing.

devinlyons:
@olleolleolle, I updated the function so it should never yield null.

skywinder:
Thanks for flagging this issue, @devinlyons!

However, the current diff has **syntax errors** — the `if not ... nil?` blocks are missing their closing `end` statements, which would cause a `SyntaxError` at load time.

If the null guard is genuinely needed, a simpler idiomatic approach would use inline guards instead of wrapping blocks:

```ruby
yield(last_response.data) if last_response.data

if parent.nil?
  until (next_one = last_response.rels[:next]).nil?
    last_response = check_github_response { next_one.get }
    yield(last_response.data) if last_response.data
  end
elsif (last = last_response.rels[:last])
  parameters = querystring_as_hash(last.href)
  last_page = Integer(parameters["page"])

  (2..last_page).each do |page|
    parent.async do
      data = check_github_response { client.send(method, user_project, *arguments, page: page, **options) }
      yield data if data
    end
  end
end
```

Would also be great to add a test case covering the nil data scenario. Could you update the PR?

Issue #1047, opened 2024-05-24
Verbose output says token is not in use


pseudotsuga-fir:
**Describe the bug**
I have put my GitHub token for the changelog generator as a variable in my environment as `CHANGELOG_GITHUB_TOKEN` as stated in the docs, however the output from the generator states the token doesn't exist: `:token=>"No token used"`.

I assume that it is actually in use because I believe I am exceeding the 50 requests per hour limit for unauthenticated requests to GitHub.

**To Reproduce**
1.  With no token provided as an argument to the generator or set in environment variables, run `github_changelog_generator`. See that verbose output states `:token=>"No token used"`.
2. Run the generator again with a token provided as an argument: `github_changelog_generator -t [YOUR_TOKEN_HERE]`. See that verbose output states: `:token=>"hidden value"`
3. Add token to environment variables as `CHANGELOG_GITHUB_TOKEN`. Run changelog generator and see that verbose output states: `:token=>"No token used"`.

**Expected behavior**
Assuming the generator is using the token but this isn't reflecting in the verbose output, I would expect the output to return: `:token=>"hidden value"`, as it does when a token is passed in as an argument.

**Additional context**
N/A

PR #1048, opened 2024-06-12
Add Token Location to Verbose Output

pseudotsuga-fir:
## What I did
Fixed issue #1047 where verbose output did not detect when token came from the `CHANGELOG_GITHUB_TOKEN` ENV variable. Updated value to specify if it came from options or ENV vars.
## How I did it
Added new method `censored_token` to `options.rb`. Method censors the token by specifying where the token came from, or if it wasn't used.
## How to verify
- Run the changelog generator with no token in ENV vars or in options. See that verbose output includes: `:token=>"No token used"`
- Run the changelog generator with token in options. See that output includes: `:token=>"Used token from options"`
- Run the changelog generator with token in ENV vars. See that output includes: `:token=>"Used token from environment variable"`
- Run the changelog generator with token in both the ENV vars and options. See that output specifies token is used from options: `:token=>"Used token from options"`


olleolleolle (approved):
Thank you, @pseudotsuga-fir!

PR #1055, opened 2024-10-30
Fix missing release branch when looking for commits

sdb9696:
The default responses per page of 30 can cause the release branch to be missed out of the query results during the check for commits in the release branch.

This causes WARNING: `merge commit was not found in the release branch or tagged git history and no rebased SHA comment was found`. https://github.com/github-changelog-generator/github-changelog-generator/issues/971

This PR sets the default query response size to 100 to reduce the likelihood of this happening. A more permanent solution could be to restrict the branches query to the required branch.

### DIFF
--- a/lib/github_changelog_generator/octo_fetcher.rb
+++ b/lib/github_changelog_generator/octo_fetcher.rb
@@ -345,7 +345,7 @@ Make sure, that you push tags to remote repo via 'git push --tags'"
     # @param [String] name
     # @return [Array<String>]
     def commits_in_branch(name)
-      @branches ||= client.branches(user_project).map { |branch| [branch[:name], branch] }.to_h
+      @branches ||= client.branches(user_project, DEFAULT_REQUEST_OPTIONS).map { |branch| [branch[:name], branch] }.to_h
 
       if (branch = @branches[name])
         commits_in_tag(branch[:commit][:sha])
--- a/lib/github_changelog_generator/octo_fetcher.rb
+++ b/lib/github_changelog_generator/octo_fetcher.rb
@@ -345,7 +345,13 @@ Make sure, that you push tags to remote repo via 'git push --tags'"
     # @param [String] name
     # @return [Array<String>]
     def commits_in_branch(name)
-      @branches ||= client.branches(user_project, DEFAULT_REQUEST_OPTIONS).map { |branch| [branch[:name], branch] }.to_h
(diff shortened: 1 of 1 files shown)


olleolleolle:
Right, 30 branches lying around is nothing. Even 100 could be very little. But offering the default options to the API request should be uncontroversial.

sdb9696:
The query also brings back deleted branches which doesn’t help. I’ll look shortly to see if there’s a parameter to control that.

sdb9696:
The latest commit iterates the pages so no 100 limit either.

sdb9696:
Hey @olleolleolle, any chance of getting this merged? I’m also happy to address any suggestions if you have any.

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -345,7 +345,13 @@ def default_branch
     # @param [String] name
     # @return [Array<String>]
     def commits_in_branch(name)
-      @branches ||= client.branches(user_project).map { |branch| [branch[:name], branch] }.to_h
+      @branches ||= lambda do
+        iterate_pages(client, "branches") do |branches|
+          branches_map = branches.map { |branch| [branch[:name], branch] }.to_h
```
olleolleolle, line 350: Ruby has a specific method for this pattern: https://rubyapi.org/3.0/o/array#method-i-to_h

```suggestion
          branches_map = branches.to_h { |branch| [branch[:name], branch] }
```

lib/github_changelog_generator/octo_fetcher.rb
```diff
@@ -345,7 +345,13 @@ def default_branch
...
+          return branches_map if branches_map[name]
+        end
+        return {}
+      end[]
```
olleolleolle, line 354: Perhaps an explicit `#call` is clearer to more readers.

```suggestion
      end.call
```

olleolleolle (approved):
I looked in detail at the individual code changes.

I think the fix is good, even with the early return, this method should not need other than the named branch, really. Any such changes are outside the scope of this repair.

sdb9696:
Thanks, I've applied the suggestions and retested the code and all works ok.

sdb9696:
Hey @olleolleolle, are you ok to merge this now?

sdb9696:
Thanks for merging @olleolleolle. Do you plan to do a release anytime soon?

PR #1068, opened 2026-03-18
Guard against nil data in iterate_pages

skywinder:
## Summary
- Skip `yield` when API response data is `nil` in `iterate_pages`
- Uses idiomatic Ruby inline guards (`yield(data) if data`) instead of wrapping blocks

Supersedes #1039 — thanks to @devinlyons for identifying the issue! The original PR had syntax errors (missing `end` keywords for `if` blocks). This PR rebases onto master and applies the fix idiomatically.

## Test plan
- [x] `ruby -c` syntax check passes
- [x] Full test suite passes (187 examples, 0 failures)
- [x] CI green across Ruby 3.0–4.0


lib/github_changelog_generator/generator/entry.rb
olleolleolle, line 20: Minor: _Could_ build a Kernel method `Options(value)` that did the above, and use it.

